PR #76, opened 2021-09-23
Add more Missing Assets Analyzers

erri120:
- for `IWeaponGetter`
- for `IStaticGetter`

both implement `IModeledGetter`

### DIFF
--- a/Mutagen.Bethesda.SkyrimAnalyzer/MissingAssetsAnalyzer.cs
+++ b/Mutagen.Bethesda.SkyrimAnalyzer/MissingAssetsAnalyzer.cs
@@ -4,28 +4,18 @@ using Microsoft.Extensions.Logging;
 using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
 using Mutagen.Bethesda.Analyzers.SDK.Errors;
 using Mutagen.Bethesda.Analyzers.SDK.Results;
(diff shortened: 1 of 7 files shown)


Mutagen.Bethesda.SkyrimAnalyzer/MissingAssetsAnalyzer.cs
```diff
@@ -4,29 +4,19 @@
...
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Plugins.Records;
 using Mutagen.Bethesda.Skyrim;
-using Mutagen.Bethesda.Skyrim.Internals;
 
 namespace Mutagen.Bethesda.SkyrimAnalyzer
 {
     [Analyzer]
-    public class MissingAssetsAnalyzer : IMajorRecordAnalyzer<IArmorGetter>, IMajorRecordAnalyzer<ITextureSetGetter>
+    public partial class MissingAssetsAnalyzer : IMajorRecordAnalyzer<IArmorGetter>, IMajorRecordAnalyzer<ITextureSetGetter>,
```
Noggog, line 14: I think we can define the `IMajorRecordAnalyzer<IArmorGetter>` on the other side of the partials, so each interface is with the related analyzer code

Noggog (approved):

PR #111, opened 2023-07-06
Cell Underscore Analyzer

Troy1010:
This adds an analyzer for issue #33

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim/CellUnderscoreAnalyzer.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/CellUnderscoreAnalyzer.cs
@@ -1,0 +1,33 @@
+﻿using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
+using Mutagen.Bethesda.Analyzers.SDK.Results;
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Skyrim;
+using Noggog;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim;
+
+public class CellUnderscoreAnalyzer : IIsolatedRecordAnalyzer<ICellGetter>
+{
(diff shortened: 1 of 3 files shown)


Mutagen.Bethesda.Analyzers.Skyrim/CellUnderscoreAnalyzer.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim;
+
+public class CellUnderscoreAnalyzer : IIsolatedRecordAnalyzer<ICellGetter>
+{
+    public static readonly TopicDefinition<string?> CellUnderscoreWrong = MutagenTopicBuilder.FromDiscussion(
+            33,
+            "EDID (Editor ID) for Cells should not have underscores in them",
+            Severity.Error)
+        .WithFormatting<string?>("Cell editor ids must not have underscores");
```
Noggog, line 15: This needs to have a `{0}` in it for the passed in string to be placed in the message.
So maybe something like 
`"Cell editor ids must not have underscores: {0}"`

Noggog (changes requested):

Mutagen.Bethesda.Analyzers.Skyrim/CellUnderscoreAnalyzer.cs
Noggog: Oh!  And the issue ID should be swapped to the ID of a discussion
https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/discussions
So need to make one over there, and grab the ID it gets.

The ID of the issue is less interesting, as we may have many issues over time that apply to this analyzer

Troy1010: I believe these are resolved in the latest commit

Noggog (approved):

PR #121, opened 2024-09-19
New Analyzers

Elscrux:
- Updated to dotnet 8
- Updated Mutagen
- Lots of new analyzers
- Added editor config
- Csv output
- Some more framework adjustments


Mutagen.Bethesda.Analyzers.Engine/Drivers/RecordFrame/RecordFrameDriver.cs
```diff
@@ -59,9 +62,9 @@ public void Drive(ContextualDriverParams driverParams)
 
             var modPath = Path.Combine(_dataDataDirectoryProvider.Path, listing.ModKey.FileName);
 
-            using var stream = File.OpenRead(modPath);
-
-            var locs = RecordLocator.GetLocations(modPath, _gameReleaseContext.Release);
+            var parsingMeta = ParsingMeta.Factory(new BinaryReadParameters(), _gameReleaseContext.Release, modPath);
```
Noggog, line 64: I think this will have to be upgraded for starfield, but we can do that later

Mutagen.Bethesda.Analyzers.Skyrim/CircularLeveledIListAnalyzer.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Mutagen.Bethesda.Analyzers.Skyrim;
+
+public partial class CircularLeveledListAnalyzer
+{
+    public IEnumerable<TopicDefinition> Topics => [CircularLeveledItem, CircularLeveledNpc, CircularLeveledSpell];
+
+    private static RecordAnalyzerResult FindCircularList<T>(T leveled, Func<T, IEnumerable<FormKey>> nestedEntriesSelector, ILinkCache linkCache)
+        where T : class, IMajorRecordGetter
+    {
+        var report = new RecordAnalyzerResult();
```
Noggog, line 15: Does this report actually accumulate anything circular?  Not seeing it get mutated anywhere

I'll probably see if i can get an analyzer unit test suite up and running ASAP, as I'd like to test the analyzers solution pretty heavily.    Get a file /w some circular lists, for ex, and see/prove if this pops

It'll also serve as a nice entry point for testing new analyzers instead of leaning on the CLI

Mutagen.Bethesda.Analyzers.Skyrim/CircularLeveledItemAnalyzer.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            if (l.Entries != null)
+            {
+                return l.Entries
+                    .Select(x => x.Data)
+                    .NotNull()
+                    .Select(x => x.Reference.FormKey);
+            }
+
+            return Array.Empty<FormKey>();
```
Noggog: I think they just came out with a feature that this can now just be `return [];`  Exciting times, haha

Mutagen.Bethesda.Analyzers.Skyrim/Contextual/ConflictingVoiceTypesAnalyzer.cs
```diff
@@ -0,0 +1,57 @@
...
+
+    public ContextualAnalyzerResult Analyze(ContextualAnalyzerParams param)
+    {
+        var result = new ContextualAnalyzerResult();
+
+        foreach (var cell in param.LinkCache.PriorityOrder.WinningOverrides<ICellGetter>())
+        {
+            if (cell.IsExteriorCell()) continue;
+
+            var npcVoiceTypes = new Dictionary<INpcGetter, FormKey>();
```
Noggog: I think what you want here for the key is `Dictionary<IFormLinkGetter<INpcGetter>, IFormLinkGetter<IVoiceTypeGetter>>`

Having the whole object type as the key means it's going to do a field-by-field comparsion for the GetHash/Equals, which is not what you're wanting here?  Making the key just the FormLink means it just compares the ID, which I think is what we're after?

Not losing the type on the value means you dont have to readd it on line 42, and it'll properly stop compiling if types were to change somehow

Mutagen.Bethesda.Analyzers.Skyrim/Record/Location/MissingFieldsAnalyzer.cs
```diff
@@ -0,0 +1,44 @@
...
+public class MissingFieldsAnalyzer : IIsolatedRecordAnalyzer<ILocationGetter>
+{
+    public static readonly TopicDefinition NoParentLocation = MutagenTopicBuilder.DevelopmentTopic(
+            "No Parent Location",
+            Severity.Suggestion)
+        .WithoutFormatting("Location has no parent location - this is likely a mistake - only top level locations should have no parent location");
+
+    public IEnumerable<TopicDefinition> Topics { get; } = [NoParentLocation];
+
+    private static readonly HashSet<FormKey> ValidTopLevelLocations =
```
Noggog: In general, i like storing `IFormLinkGetter<TheType>`.   Just makes code more clear of what we're working with, imo

Are these used at all? Looks not referenced... or is this just a WIP analyzer?

Is there a way we can figure out a top level location by analyzing the data?  Or is it just arbitrary?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Key/WeightValueAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+    public static readonly TopicDefinition<float> WeightNotZero = MutagenTopicBuilder.DevelopmentTopic(
+            "Weight Not Zero",
+            Severity.Suggestion)
+        .WithFormatting<float>("Key weight {0} is not zero");
+
+    public static readonly TopicDefinition<uint> ValueNotZero = MutagenTopicBuilder.DevelopmentTopic(
+            "Value Not Zero",
+            Severity.Suggestion)
+        .WithFormatting<uint>("Key value {0} is not zero");
+
```
Noggog, line 18: Hmm, feels like an opinion analyzer.    Definitely the "core" set of analyzers I think I want to have only be errors/anomalies.  I think in general, Suggestion as a severity will be rare.  Here were the notes from years ago about them:
> Suggestions currently have stuff like:
Conditions can be simplified.   Not a warning.  Not causing bugs.  But we suggest improving it
New but deleted records.   Someone made an item then deleted it.  Just fluff.   Not really a "warning", per se, but we can suggest just removing the totally unused thing

The idea was they're bringing up things that aren't problems, necessarily, but aren't subjectively opinionated, either.

But this does bring back up the idea of how to sideload analyzers, as that's where i think these styles of "suggestion" analyzers would likely live?

To me, this sort of opinion logic "this is how -I- like my game" generally feels more appropriate to being a synthesis patcher.   Just make a patcher that whitewashes it all to be zero to your liking (or even 0.1 as a setting)

Maybe a side analyzer could be appropriate during dev work.  "within the XYZ dev team project, we want to enforce keys are always weightless" and so you'd have an analyzer.  Synthesis patch isn't ideal there, as you'd want to modify the original mod if any were detected

Still, that feels like it's something special to XYZ dev team, not the wide populace

Mutagen.Bethesda.Analyzers.Skyrim/Record/Ingredient/MissingFieldsAnalyzer.cs
```diff
@@ -0,0 +1,67 @@
...
+            "No Pickup Sound",
+            Severity.Suggestion)
+        .WithoutFormatting("Ingredient has no pickup sound");
+
+    public static readonly TopicDefinition NoPutDownSound = MutagenTopicBuilder.DevelopmentTopic(
+            "No Put Down Sound",
+            Severity.Suggestion)
+        .WithoutFormatting("Ingredient has no put down sound");
+
+    public static readonly TopicDefinition<int> NotFourEffects = MutagenTopicBuilder.DevelopmentTopic(
```
Noggog, line 19: Now this is more of a suggestion I was imagining.  Since it's sort of "in the engine" that ingredients have 4 effects... it's abnormal to not have 4?  Seems a bit less opinion than weight values, haha.  

Might even be a warning?  What happens if someone only defines 3?  or does 5?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Ingestible/MissingFieldsAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
+﻿using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
+using Mutagen.Bethesda.Analyzers.SDK.Results;
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Skyrim;
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Ingestible;
+
+public class MissingFieldsAnalyzer : IIsolatedRecordAnalyzer<IIngestibleGetter>
+{
+    public static readonly TopicDefinition EmptyEffectList = MutagenTopicBuilder.DevelopmentTopic(
```
Noggog, line 9: Feels more Suggestion tier?   Here were the descriptions of the levels I made ages ago:

> Critical/CTD:   Absolute fundamental problem that crashes or freezes the game
Error:  Something's going to be wrong, has a known negative side effect, but still playable.  Missing textures, etc
Warning:  Something's odd or wrong, but doesn't have any known negative side effects.  Might be upgraded to error if one was discovered
Suggestion:  Optimizations, small improvements or other suggestions.

Where warning was "might be an error, but we're not sure yet, haha"

This is "oh, this looks maybe vestigial or incomplete? Might want to finish it off?", which feels like a suggestion caliber, to me.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Faction/VendorHourAnalyzer.cs
```diff
@@ -0,0 +1,37 @@
...
+using Mutagen.Bethesda.Analyzers.SDK.Results;
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Skyrim;
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Faction;
+
+public class VendorHourAnalyzer : IIsolatedRecordAnalyzer<IFactionGetter>
+{
+    public static readonly TopicDefinition<ushort, ushort> WrongVendorHourOrder = MutagenTopicBuilder.DevelopmentTopic(
+            "Wrong Vendor Hour Order",
+            Severity.Warning)
```
Noggog: I think this gets upgraded to Error, imo

>  Something's going to be wrong, has a known negative side effect, but still playable

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Topic/TooLongAnalyzer.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Dialog.Topic;
+
+public class TooLongAnalyzer : IIsolatedRecordAnalyzer<IDialogTopicGetter>
+{
+    private const int DialogPromptLengthLimit = 80;
+    public static readonly TopicDefinition<string, int> TopicPromptTooLong = MutagenTopicBuilder.DevelopmentTopic(
+            "Topic Prompt Too Long",
+            Severity.Suggestion)
+        .WithFormatting<string, int>("Topic prompt '{0}' is {1} longer than the recommended limit " + DialogPromptLengthLimit);
+
```
Noggog, line 14: Where does this recommended limit come from?  Feels like something that might be upgraded to Error, potentially

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Responses/SkillCheckAnalyzer.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    public static readonly TopicDefinition<string> NonPlayerSkillCheck = MutagenTopicBuilder.DevelopmentTopic(
+            "Non-Player Skill Check",
+            Severity.Warning)
+        .WithFormatting<string>("Skill check in dialog are not checked on the player but on {0} - this is usually a sign of a mistake");
+
+    public static readonly TopicDefinition<string> NonGlobalSkillCheck = MutagenTopicBuilder.DevelopmentTopic(
+            "Non-Global Skill Check",
+            Severity.Suggestion)
+        .WithFormatting<string>("Skill check in dialog doesn't use global to evaluate skill level but {0} - this is usually a sign of a mistake");
```
Noggog: Any reason you marked one as warning, the other as a suggestion?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Responses/PersuasionAmuletConditionAnalyzer.cs
```diff
@@ -0,0 +1,50 @@
...
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Skyrim;
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Dialog.Responses;
+
+public class PersuasionAmuletConditionAnalyzer : IIsolatedRecordAnalyzer<IDialogResponsesGetter>
+{
+    public static readonly TopicDefinition MissingCreatedObject = MutagenTopicBuilder.DevelopmentTopic(
+            "Missing Amulet of Articulation Condition",
+            Severity.Suggestion)
+        .WithoutFormatting("Persuasion check is missing auto pass condition when Amulet of Articulation is equipped");
```
Noggog, line 12: Woof.  Breaking my brain.  Definition subjective/opinionated to a specific game setup, but still is helpful and interesting, haha.

Like, what if it's Enderal or some other game that doesn't even have the Amulet of Articulation?   

And why is this item special?   What about other items that are from mods that expect similar things?  Why no analyzers for them?    Sideloads to the rescue?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Responses/InvalidSayOnceAnalyzer.cs
```diff
@@ -0,0 +1,39 @@
...
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Skyrim;
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Dialog.Responses;
+
+public class InvalidSayOnceAnalyzer : IContextualRecordAnalyzer<IDialogResponsesGetter>
+{
+    public static readonly TopicDefinition<string> InvalidSayOnce = MutagenTopicBuilder.DevelopmentTopic(
+            "Invalid SayOnce",
+            Severity.Warning)
+        .WithFormatting<string>("Dialog is say once although it's quest {0} is not Start Game Enabled");
```
Noggog: Does this have a notable negative impact somehow?  if so, upgrade to Error

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Responses/InconsistentCharactersAnalyzer.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    public static readonly TopicDefinition<string, string> PromptInconsistentCharacters = MutagenTopicBuilder.DevelopmentTopic(
+            "Prompt Has Inconsistent Characters",
+            Severity.Suggestion)
+        .WithFormatting<string, string>("Response {0} contains characters {1} which are not usually used in dialog");
+
+    public static readonly TopicDefinition<string, string> ResponseInconsistentCharacters = MutagenTopicBuilder.DevelopmentTopic(
+            "Response Has Inconsistent Characters",
+            Severity.Suggestion)
+        .WithFormatting<string, string>("Response {0} contains characters {1} which are not usually used in dialog");
```
Noggog: Why are these not allowed?  Does it cause issues?

Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        return record switch
+        {
+            ICameraPathGetter cameraPath => cameraPath.Conditions,
+            IConstructibleObjectGetter constructibleObject => constructibleObject.Conditions,
+            IDialogResponsesGetter dialogResponses => dialogResponses.Conditions,
+            IFactionGetter faction => faction.Conditions,
+            IIdleAnimationGetter idleAnimation => idleAnimation.Conditions,
+            ILoadScreenGetter loadScreen => loadScreen.Conditions,
+            IMagicEffectGetter magicEffect => magicEffect.Conditions,
```
Noggog, line 66: I would like to find a more automatically maintaining system for ones like this.  As games are being developed (Starfield for, ex), these lists will be hard to keep updated to the changing definitions

If this gets too bad, we can either:
- Add a `GetConditions` to mutagen
- Maybe even make a source generator.  Something like `GetFieldsOfType<IConditionGetter>(rec)`, where the source generator would analyze the fields of the record type and make a `GetConditions` style function on the fly.

Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
```diff
@@ -0,0 +1,82 @@
...
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Skyrim;
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record;
+
+public class ConditionAnalyzer : IIsolatedRecordAnalyzer<ISkyrimMajorRecordGetter>
+{
+    private static readonly TopicDefinition<string?> InvalidConditionReference = MutagenTopicBuilder.DevelopmentTopic(
+            "Condition Runs on Null Reference",
+            Severity.Error)
+        .WithFormatting<string?>("Condition {0} runs on reference, but reference is null");
```
Noggog, line 12: I think most topics of type error should have some notes on what the known problems they cause in-game are.

Normally this would go in the Discussion topic on github.  But while it's just a `DevelopmentTopic`, we need to keep the notes somewhere.  If it was initiated by an issue, we'd keep them there.. but for now maybe add comments?

Otherwise, when we eventually "upgrade" it off `DevelopmentTopic`, I'm not gonna know the driving original details of why it was an error

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/VendorKeywordAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+public class VendorKeywordAnalyzer : IIsolatedRecordAnalyzer<IArmorGetter>
+{
+    public static readonly TopicDefinition<string, string> ArmorMissingVendorKeyword = MutagenTopicBuilder.DevelopmentTopic(
+            "Armor is missing Vendor Keyword",
+            Severity.Suggestion)
+        .WithFormatting<string, string>("Missing vendor keyword {0} but has {1}");
+
+    public IEnumerable<TopicDefinition> Topics => [ArmorMissingVendorKeyword];
+
+    private sealed record NamedFormLink(FormLink<IKeywordGetter> FormLink, string EditorID);
```
Noggog: We do have `IMajorRecordIdentifier`, but that loses the <IKeywordGetter> typing

I might add `IMajorRecordIdentifier<TMajorType> : IFormLinkGetter<TMajorType>` which could fully take `NamedFormLink`'s place
https://github.com/Mutagen-Modding/Mutagen/issues/556

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/VendorKeywordAnalyzer.cs
```diff
@@ -0,0 +1,97 @@
...
+using Mutagen.Bethesda.Skyrim;
+using Noggog;
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Armor;
+
+public class VendorKeywordAnalyzer : IIsolatedRecordAnalyzer<IArmorGetter>
+{
+    public static readonly TopicDefinition<string, string> ArmorMissingVendorKeyword = MutagenTopicBuilder.DevelopmentTopic(
+            "Armor is missing Vendor Keyword",
+            Severity.Suggestion)
+        .WithFormatting<string, string>("Missing vendor keyword {0} but has {1}");
```
Noggog: Asking more questions, just to figure out how things fit into categories better.

Does this cause an issue in game?  Or is this just typically the pattern that things should have to make sense?  What happens when an armor is missing the vendor keyword?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/RaceAddonsAnalyzer.cs
```diff
@@ -0,0 +1,75 @@
...
+public class RaceAddonsAnalyzer : IContextualRecordAnalyzer<IArmorGetter>
+{
+    public static readonly TopicDefinition<string?> ArmorMissingRaceAddons = MutagenTopicBuilder.DevelopmentTopic(
+            "Armor is missing race addons",
+            Severity.Warning)
+        .WithFormatting<string?>("Missing race addon for race: {0}");
+
+    public IEnumerable<TopicDefinition> Topics => [ArmorMissingRaceAddons];
+
+    private static readonly List<FormKey> DefaultPlayerRaces =
```
Noggog: `HashSet<IFormLinkGetter<IRaceGetter>>` more appropriate here

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/RaceAddonsAnalyzer.cs
```diff
@@ -0,0 +1,75 @@
...
+                missingRaces.Remove(additionalRace.FormKey);
+            }
+            missingRaces.Remove(addon.Race.FormKey);
+        }
+
+        var result = new RecordAnalyzerResult();
+
+        foreach (var race in missingRaces)
+        {
+            if (param.LinkCache.TryResolve<IRaceGetter>(race, out var r))
```
Noggog: Hmm.  The resolution here to get the EditorID feels wrong, as the editor ID is just for "convenience" in the log message to include the EditorID in addition to the FormKey?

The problem is that if the Race cant be found, this topic wont be added.  I'm guessing that's not desired?

I think instead, we'll want to upgrade the engine.   We should be able to provide a topic here with just the FormKey of the race.  The engine can do the lookup and decorate with the EditorID when appropriate?  That way the topic gets reported either way, but the EditorID is included when it can be, and also each analyzer doesn't need to remember to do anything special to get the perk

https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/issues/123

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/RaceAddonsAnalyzer.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                result.AddTopic(RecordTopic.Create(
+                    armor,
+                    ArmorMissingRaceAddons.Format(r.EditorID),
+                    x => x.EditorID));
```
Noggog: I think `x => x.EditorID` is a misuse of this input.

The purpose of `memberExpression` when making a topic is to inform the engine what member of the original record of interest it should watch is.

For example.  In a simpler analyzer, like say one that makes sure Npc weights are above zero, or something... it would put `x => x.Weight`
- Engine runs analyzer on Npc A
- Analyzer finds that weight was < 0, reports topic, /w a `x => x.Weight` memberExpression
- File gets updated
- Engine checks if NpcA's weight field got changed
- It was!  Reruns the analyzer, finds that it's now 1, doesn't re-report topic

This is the "dream" of an optimized but live engine that is able to intelligently rerun analyzers only when appropriate.

So here, I think this is being misused, as the Armor's EditorID is not involved in determining if the analyzer should be rerun.  In fact... i think this analyzer is complex enough that it might not be able to specify a field to watch in this manner?  

I think we'll want to also upgrade the API to allow for "multiple members" to be watched easily, and maybe even a custom func comparer alternative.  And for the short term.. NO member as well (as right now it requires one to be specified)

https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/issues/124

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/KeywordSlotsAnalyzer.cs
```diff
@@ -0,0 +1,76 @@
...
+            ]
+        };
+
+        // Armor type independent conditions
+        conditions.AddRange([
+            (BipedObjectFlag.Shield, FormKeys.SkyrimSE.Skyrim.Keyword.ArmorShield, nameof(FormKeys.SkyrimSE.Skyrim.Keyword.ArmorShield)),
+            (BipedObjectFlag.Circlet, FormKeys.SkyrimSE.Skyrim.Keyword.ClothingCirclet, nameof(FormKeys.SkyrimSE.Skyrim.Keyword.ClothingCirclet)),
+            (BipedObjectFlag.Amulet, FormKeys.SkyrimSE.Skyrim.Keyword.ClothingNecklace, nameof(FormKeys.SkyrimSE.Skyrim.Keyword.ClothingNecklace)),
+            (BipedObjectFlag.Ring, FormKeys.SkyrimSE.Skyrim.Keyword.ClothingRing, nameof(FormKeys.SkyrimSE.Skyrim.Keyword.ClothingRing)),
+        ]);
```
Noggog: These types of lists i think can be made statically ahead of time.  No need to remake this list every time we want to run this analyzer?  Just need to maybe make the two alternatives ahead of time, and then choose the right list here

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/WorldspaceExtension.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Mutagen.Bethesda.Plugins.Cache;
+using Mutagen.Bethesda.Skyrim;
+namespace Mutagen.Bethesda.Analyzers.Skyrim;
+
+public static class WorldspaceExtension
+{
+    public static IEnumerable<ILocationGetter> GetWorldLocations(this IWorldspaceGetter world, ILinkCache linkCache)
+    {
+        var worldLocations = new HashSet<ILocationGetter>();
```
Noggog: Either will want to be `List<ILocationGetter>` or `HashSet<IFormLinkGetter<ILocationGetter>>`

This right now is gonna hash all the fields of the location, which is not desired, i dont think

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/LocationExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+        var isSettlement = location.Keywords.Any(k =>
+            k.FormKey == FormKeys.SkyrimSE.Skyrim.Keyword.LocTypeHabitation.FormKey
+            || k.FormKey == FormKeys.SkyrimSE.Skyrim.Keyword.LocTypeHabitationHasInn.FormKey
+            || k.FormKey == FormKeys.SkyrimSE.Skyrim.Keyword.LocTypeFarm.FormKey
+            || k.FormKey == FormKeys.SkyrimSE.Skyrim.Keyword.LocTypeCity.FormKey
+            || k.FormKey == FormKeys.SkyrimSE.Skyrim.Keyword.LocTypeHouse.FormKey
+            || k.FormKey == FormKeys.SkyrimSE.Skyrim.Keyword.LocTypeSettlement.FormKey
+            || k.FormKey == FormKeys.SkyrimSE.Skyrim.Keyword.LocTypePlayerHouse.FormKey
+            || k.FormKey == FormKeys.SkyrimSE.Skyrim.Keyword.LocTypeInn.FormKey
+            || k.FormKey == FormKeys.SkyrimSE.Skyrim.Keyword.LocTypeStore.FormKey);
```
Noggog: Can probably shove all these into a `HashSet<IFormLinkGetter<IKeywordGetter>>` (constructed statically) and do a contains

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/CellExtensions.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System.Collections.Concurrent;
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Plugins.Cache;
+using Mutagen.Bethesda.Skyrim;
+namespace Mutagen.Bethesda.Analyzers.Skyrim;
+
+public static class CellExtensions {
+    private static readonly ConcurrentDictionary<FormKey, IReadOnlySet<ILocationGetter>> CellLocationCache = new();
```
Noggog: Dont think this static dictionary will be safe to do.  Since its parent functions seem unused, let's remove for now?

Mutagen.Bethesda.Analyzers.Skyrim/Mutagen.Bethesda.Analyzers.Skyrim.csproj.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/CodeInspection/NamespaceProvider/NamespaceFoldersToSkip/=extensions/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
```
Noggog: Not familiar with this file.  What's it doing for us?

Noggog (changes requested):
Thanks for the PR!  It's a great start to getting the analyzers project seeded!

Don't get too overwhelmed by the comments.  Was working through some design waffling while making them.    I think generally since most of the analyzers are marked `DevelopmentTopic`, we can just evaluate them in greater detail as we go to "upgrade" them to official stamped topics

Overall though, this PR's convinced me a bit more than we need sideloading.   So ill be thinking on that more

There's a few comments that could be adjusted before we merge this in.. but the bulk is just discussion/musings

Elscrux:
> Thanks for the PR! It's a great start to getting the analyzers project seeded!
> 
> Don't get too overwhelmed by the comments. Was working through some design waffling while making them. I think generally since most of the analyzers are marked `DevelopmentTopic`, we can just evaluate them in greater detail as we go to "upgrade" them to official stamped topics
> 
> Overall though, this PR's convinced me a bit more than we need sideloading. So ill be thinking on that more
> 
> There's a few comments that could be adjusted before we merge this in.. but the bulk is just discussion/musings

Thanks a lot for working through this massive PR. I appreciate all the suggestions!

Mutagen.Bethesda.Analyzers.Skyrim/CircularLeveledIListAnalyzer.cs
Elscrux: Yeah seems like I forgot to add that properly. This was one of the few analyzers I didn't test properly because I didn't have any plugin with this issue at hand.

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/CellExtensions.cs
Elscrux: Yep, was and old optimization that I forgot to remove when copying over my code. The function itself will be used in future tests that I'll add so I'll keep that for now but the dictionary will go.

Mutagen.Bethesda.Analyzers.Skyrim/Contextual/ConflictingVoiceTypesAnalyzer.cs
Elscrux: We'll still want to keep the editor id of the npcs at least so I'll downgrade the dictionary key to `IMajorRecordIdentifier`. Upgrading the value to `IFormLinkGetter<IVoiceTypeGetter>` seems good

Mutagen.Bethesda.Analyzers.Skyrim/Record/Location/MissingFieldsAnalyzer.cs
Elscrux: Yeah I forgot to integrate the valid top level locations list.
This is pretty much just arbitrary. There are some valid ones like the top level location for the Skyrim worldspace (could maybe be found by removing any locations that are assigned to locations, I'll add that). Some are sort of hardcoded or some are part of daedric realms which don't really belong to any bigger location. In the end everything in the real world will be part of tamriel or nirn so that should always have a parent location. There are also some technical locations like the `PersistAll` which should always be ignored.
I'm not super happy with this whitelist approach here, and people would probably would like to be able to configure that stuff on their own potentially. I just included the most universal ones here for now. There might still be some false positives from the analyzer though - that'd be up to the user to decide currently.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Key/WeightValueAnalyzer.cs
Elscrux: I think this specific example is just enforcing consistency with Skyrim. All but one key in the game don't have a weight.
But yeah I definitely have some more opinionated analyzers in there which is why I pushed for sideloading of analyzers.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Ingredient/MissingFieldsAnalyzer.cs
Elscrux: When there are just 3 effects it can use which is inconsistent. When it's 5 or more it just ignores all after the 4th.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Topic/TooLongAnalyzer.cs
Elscrux: It's a limit set in the CK by default. But it can be overridden via ini tweak. Technically more is possible but it is a recommendation by Bethesda so it fits on the screen, is consistent and so on.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Responses/SkillCheckAnalyzer.cs
Elscrux: The non-player skill check is usually an error that causes unexpected behavior but we're not sure so it's a warning. The global thing is demanded by convention but technically it also works when it's a float. Just doesn't get affected by perks that make speech skills easier by decreasing the global.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Responses/PersuasionAmuletConditionAnalyzer.cs
Elscrux: I think Enderal (seeing as it has it's own Mutagen game version) is quite separate and we can assume that people play Skyrim here. But for sure there could be a category of analyzers in the future that are specifically for these kinds of things.
This is special because it is a precedent set by Bethesda. Other mods that add this sort of stuff would need to add their stuff to the formlist which would make them automatically compatible. If they would like to add another way to auto pass skill checks then they would need to adjust the whole base game to be consistent which is not feasible from a compatibility standpoint so nobody does that.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Responses/InconsistentCharactersAnalyzer.cs
Elscrux: It's a bit opinionated because these are characters that my project uses to mark placeholder lines and notes how dialogue should behave which I would like to report when they're still in game. In general though, these characters are never used in Skyrim, only in unfinished content and one regular case which is an anomaly. Using it would definitely be inconsistent with Skyrim's typical writing.

Here are two examples of cut content where it's used
> Response Join the forces mustering at the [CAPITAL] (infos conditions on location go here)
> Response Join the forces mustering outside of [CAPITAL]. (Infos conditioned on Capital location go here)

Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
Elscrux: Would be good to have a better way to get these, yeah!

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/VendorKeywordAnalyzer.cs
Elscrux: Yeah at least for this example with the form keys libraries it would be nice to access the editor as field as well although I hope this doesn't further inflate their dll size that much.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/VendorKeywordAnalyzer.cs
Elscrux: It means that you can't sell the armor in game. The way vendors function is fully keyword drive. There is a vendor X who buys all items that have keyword Y. So only in very few cases this shouldn't be the case.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/RaceAddonsAnalyzer.cs
Elscrux: Yeah I agree that would be good

Mutagen.Bethesda.Analyzers.Skyrim/Mutagen.Bethesda.Analyzers.Skyrim.csproj.DotSettings
Elscrux: This marks the extensions folder to be no namespace provider, meaning the namespace will not automatically be set up to follow the folder structure with `namespace Mutagen.Bethesda.Analyzers.Skyrim.Extensions;` and instead it will stay at `namespace Mutagen.Bethesda.Analyzers.Skyrim;` meaning all namespaces in the project can use the extensions without adding any `using` statements.

Mutagen.Bethesda.Analyzers.Skyrim/Contextual/ConflictingVoiceTypesAnalyzer.cs
Noggog: Not sure what you mean by `We'll still want to keep the editor id of the npcs`?

Are you saying you want to keep the EditorID as part of the key that determines collisions?  
I don't think swapping the key to `IMajorRecordIdentifier` will be the fix.  It won't reduce the things that are considered when determining key collisions, it will just restrict what can go into the key.  But whatever is put in there still uses its normal equality/hashcode... so it'll be essentially the same as it was before, but now you could put a weapon in there as well as a key.

I think either you'll want to do the `IFormLinkGetter<INpcGetter>` as a key, which requires `npc.ToFormLink()` when putting it in.  That's the fundamental difference, as the `ToFormLink` actually makes a FormLink class, and so the FormLink.Equals will be what's used for collisions.

But I'm remembering now that we have something even better:
`new Dictionary<INpcGetter, IFormLinkGetter<IVoiceTypeGetter>>(MajorRecord.FormKeyEqualityComparer)`
This lets you put the full object in as the key (for ease), but instructs the dictionary exactly how to consider collisions to just be the FormKey.  (I dont think EditorIDs are worth considering?)

Elscrux: It was mostly needed to include the editor id to keep it around for the report but yeah you're totally right that this still has issues as a key. As I've added the possibility to pass form links into the topic which will be resolved when they are reported, I was able to fully remove this part from the analyzer.

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/FactionExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Mutagen.Bethesda.Skyrim;
+namespace Mutagen.Bethesda.Analyzers.Skyrim;
+
+public static class FactionExtensions
+{
+    public static bool IsVendor(this IFactionGetter faction)
+    {
+        return (faction.Flags & Faction.FactionFlag.Vendor) != 0;
```
Noggog, line 8: We should use HasFlag
`return faction.Flags.HasFlag(Faction.FactionFlag.Vendor);`

Mutagen.Bethesda.Analyzers.Skyrim/Mutagen.Bethesda.Analyzers.Skyrim.csproj.DotSettings
Noggog: Hmm, I dont necessarily mind the extensions living in their own namespace to match the folder they're in.   IDEs will import them easily enough.    If we dont want to have to do that, I think I'd prefer adding implicit namespaces to the project instead?  Similar result, but without a Jetbrains specific file

https://devblogs.microsoft.com/dotnet/welcome-to-csharp-10/

Mutagen.Bethesda.Analyzers.Engine/Reporting/CsvDropbox.cs
```diff
@@ -0,0 +1,48 @@
...
+    private static string BuildLine(ITopic topic, IModGetter? sourceMod, IMajorRecordGetter? majorRecord)
+    {
+        return $"""
+        "{topic.TopicDefinition}","{sourceMod?.ModKey.ToString()}","{majorRecord?.FormKey.ToString()}","{majorRecord?.EditorID}","{topic.FormattedMessage}"
+        """;
+    }
+
+    private void Append(string line)
+    {
+        if (_reportOutputConfiguration.OutputFilePath is null)
```
Noggog: I think this short circuit is too late.   We're doing a lot of work to get to the line string, just to toss it out the window if the feature isn't even active.

I'd go so far as to say we should only register CsvDropbox as a decorator if the feature is active.    In `RunAnalyzers.cs`, we're building the container and know if the output file is active, and can optionally register at that point

Mutagen.Bethesda.Analyzers.Cli/Modules/RunAnalyzerModule.cs
```diff
@@ -10,6 +10,9 @@ protected override void Load(ContainerBuilder builder)
     {
         builder.RegisterDecorator<MinimumSeverityFilter, IReportDropbox>();
         builder.RegisterDecorator<SeverityAdjuster, IReportDropbox>();
+        builder.RegisterDecorator<CsvDropbox, IReportDropbox>();
+        builder.RegisterDecorator<TopicListJoin, IReportDropbox>();
+        builder.RegisterDecorator<TopicEnricher, IReportDropbox>();
```
Noggog, line 22: Let's back off the TopicListJoin/TopicEnricher card/issue for now and do in another PR

Mutagen.Bethesda.Analyzers.SDK/Topics/ITopic.cs
```diff
@@ -5,4 +5,5 @@ public interface ITopic
     TopicDefinition TopicDefinition { get; }
     string FormattedMessage { get; }
     Severity Severity { get; set; }
+    IEnumerable<object?> Items { get; }
```
Noggog, line 8: Don't know if I'm on board of disrupting the interface like this yet.   I'd want to think about the EditorID enrichment card in isolation a bit to know if this was worth it or not... but I'd rather not hold up this PR for it.   We can always mix in EditorID enrichment later.. it's not a blocker for any of the rest of the content I dont think

Can we back out the changes for this card for now?
- This `Items` field
- FuncRecordTopic class
- Topic enrichers

Noggog (changes requested):

Noggog:
Also, from the #118 discussion

> Gotcha. Might be nice to not have it so random. Maybe the TopicPrefix can be DEV
Maybe we just modify TopicId to take in a string id, where the DevelopmentTopic passes in a string, but the FromDiscussion passes the int

Mutagen.Bethesda.Analyzers.Engine/Reporting/CsvDropbox.cs
Elscrux: Yeah sounds good. Nullability will still be an issue here though as `_reportOutputConfiguration.OutputFilePath` is currently nullable coming from `RunAnalyzersCommand`, so I had to rework some stuff.

Elscrux:
> Hmm, I dont necessarily mind the extensions living in their own namespace to match the folder they're in. IDEs will import them easily enough. If we dont want to have to do that, I think I'd prefer adding implicit namespaces to the project instead? Similar result, but without a Jetbrains specific file
> 
> https://devblogs.microsoft.com/dotnet/welcome-to-csharp-10/

Sounds good! Wasn't aware this is a JetBrains specific thing

Mutagen.Bethesda.Analyzers.SDK/Topics/ITopic.cs
Elscrux: I personally think this is quite essential to have analyzers that you can properly work with. I refactored a lot of code so we pass form links which makes the development experience a lot better when you don't have to care about getting any editor ids. If you think there is a good way to handle this differently let me know! I agree it's a bit hacky right now. Maybe we could add a new type which combines something like `FormattedMessage` and `Items` which are both concerned with the resulting message sent to the dropbox. I personally wouldn't mind waiting a bit with the PR while we're figuring this out. Feels incomplete otherwise.

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/FactionExtensions.cs
Elscrux: I know we have `HasFlag` but I thought for extension utility calls like this it would be preferred to use the faster bitwise operation? Probably doesn't matter very much but as this can be used a lot in various scenarios it might be worth sticking to this?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/VendorKeywordAnalyzer.cs
Noggog: Ah, didn't think about that.    https://github.com/Mutagen-Modding/Mutagen.Bethesda.FormKeys/issues/5

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/FactionExtensions.cs
Noggog: Fair.  I'll see if i can add some extension function to get the best of both

Mutagen.Bethesda.Analyzers.Cli/Overrides/CustomDataDirectoryProvider.cs
```diff
@@ -0,0 +1,9 @@
+﻿using Mutagen.Bethesda.Environments.DI;
+using Noggog;
+namespace Mutagen.Bethesda.Analyzers.Cli.Overrides;
+
+public class CustomDataDirectoryProvider(string path) : IDataDirectoryProvider
```
Noggog: There's a class `DataDirectoryInjection` that already exists

Mutagen.Bethesda.Analyzers.Cli/Overrides/CustomEnabledPluginListingsProvider.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Plugins.Order;
+using Mutagen.Bethesda.Plugins.Order.DI;
+namespace Mutagen.Bethesda.Analyzers.Cli.Overrides;
+
+public class CustomEnabledPluginListingsProvider(string path) : IEnabledPluginListingsProvider
```
Noggog: Hmm, so we're taking in a custom DataFolder, and using that to drive a replacement for Plugins.txt, just assuming any mod that is present is one to watch?

That's a bit deceiving as the CLI API suggests it's just setting the datafolder -> not that it's overriding the plugins.txt completely as well

What's the reasoning for wanting to override this in this manner?  For the continuous integration, i imagine?  If so, maybe we add another parameter `--UseDataFolderForLoadOrder` or something

------------

Alternatively, I think the analyzer project in general shouldn't be too concerned with plugins.txt overall.  Like, when a user opens up the analyzer UI, it should just analyze all the files in the data folder.  It should also consider plugins.txt for some specific analyzers, but that's more of an afterthought than a required input.

So, perhaps instead of adding `--UseDataFolderForLoadOrder` we add `--WithNoLoadOrder` and omit the plugins.txt concepts entirely (potentially turning off a few analyzers dedicated to analyzing load order interactions)

Noggog (changes requested):

Mutagen.Bethesda.Analyzers.Cli/Overrides/CustomEnabledPluginListingsProvider.cs
Elscrux: Yeah I need it for CI. `--WithNoLoadOrder` wouldn't really work for my case when it would turn of the load order analyzers because I'd still wanted them in there, just in this different environment. Having an implicit `--UseDataFolderForLoadOrder` could be good I think.
Any analyzer UI would have their own registration so I think this would be entirely separate from the cli stuff.

Noggog (approved):
Damn!  Now that's some seed content!

PR #138, opened 2024-09-27
130 evolve report pipeline

### LINKED ISSUE #130: Evolve Report Pipeline
On the analyzer end:
`IReportDropbox` what's passed to the analyzers to offer a simple drop in API.  They just need to give us their topic and we'll handle the rest

On the other end:
`IReportHandler` A specialized reporting endpoint.  There might be multiple of these active at once.  Say, `ConsoleReportHandler` and `CsvReportHandler`.  They take the topics, with their message + extra items, and decide how to format them for their end target.

In the middle, near the analyzer side:
`ReportDropboxEngineHelper : IReportDropbox` this would be an engine internal helper class.   IReportHandler might need lots of info like ILinkCache.  Analyzers don't want to worry about that (some might not even have it).   This would be the guy in between coordinating.  It would intercept any drops by the analyzer, and pass them to the handlers /w all the extra needed bits, like the currently active link cache.

Who does the EditorID resolve and swap out?  Could go a few places.  Maybe we define a section in the middle before it hits the end handlers that does some transformation.   EngineHelper gives it to the `IReportTransformer` which might be decorated a bunch, but eventually spits out the transformed item to be passed to the handler?

Noggog:
# IReportHandler
Gets all reports in their final form and decides where they go
- CsvReportHandler
- ConsoleReportHandler

# IReportDropbox
Thing that the engine funnels reported topics into.  Gets enriched with linkcache.   Is meant to be decorated to create a "pipeline"
- MinimumSeverityFilter
- SeverityAdjuster
- EditorIdEnricher
- (disabled for now) DisallowedParametersChecker
- (capstone) PassToHandlerReportDropbox

Refactored some topic records to new C# standards


Noggog:
Oh, for the record, I have -no- idea if this currently wires up and runs still, haha.  I should try real quick (havent actually run analyzers yet since all the activity)

Mutagen.Bethesda.Analyzers.Engine/Reporting/Handlers/CsvReportHandler.cs
```diff
@@ -1,42 +1,46 @@
 ﻿using Mutagen.Bethesda.Analyzers.SDK.Topics;
 using Mutagen.Bethesda.Plugins.Records;
-namespace Mutagen.Bethesda.Analyzers.Reporting;
 
-public class CsvDropbox : IReportDropbox
+namespace Mutagen.Bethesda.Analyzers.Reporting.Handlers;
+
+public record CsvInputs(string OutputFilePath);
```
Elscrux, line 6: I don't think this is being registered at the moment, should be added next to the registration of `CsvReportHandler`

Mutagen.Bethesda.Analyzers.Engine/Reporting/Drops/EditorIdEnricher.cs
```diff
@@ -0,0 +1,67 @@
...
+        ITopic topic)
+    {
+        return topic.WithFormattedTopic(topic.FormattedTopic.Transform(parameters, Selector));
+    }
+
+    private object LinkResolver(
+        ReportContextParameters parameters,
+        IFormLinkGetter link)
+    {
+        if (parameters.LinkCache.TryResolve(link.FormKey, link.Type, out var record)
```
Elscrux: There's also the `TryResolveIdentifier` method. Not sure if that's more optimized in some way? Seems appropriate though.

Mutagen.Bethesda.Analyzers.Engine/Reporting/Drops/DisallowedParametersChecker.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        return topic.WithFormattedTopic(
+            topic.FormattedTopic.Transform(parameters, Checker));
+    }
+
+    private object? Checker(
+        ReportContextParameters parameters,
+        object? item)
+    {
+        if (item is IEnumerable)
```
Elscrux: I would probably make this more specific. `string` is also an enumerable of `char`. Maybe `ICollection` or so?

Mutagen.Bethesda.Analyzers.Engine/Reporting/TopicListJoin.cs
Elscrux, line 1: I think we'll still want to get this back when we're adding meta data to attach to topics

Mutagen.Bethesda.Analyzers.Engine/Reporting/Drops/EditorIdEnricher.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        return link;
+    }
+
+    private object? Selector(
+        ReportContextParameters parameters,
+        object? item)
+    {
+        if (item is IFormLinkGetter link)
```
Elscrux, line 61: Enriching nested structures like dictionaries or lists like the previous implementation did should still be a thing for future additions like the meta data for topics I feel like?

Mutagen.Bethesda.Analyzers.Engine/Reporting/Drops/IReportDropbox.cs
```diff
@@ -0,0 +1,17 @@
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Plugins.Records;
+
+namespace Mutagen.Bethesda.Analyzers.Reporting.Drops;
+
+public interface IReportDropbox
```
Elscrux, line 6: Somehow feels a bit strange to have an exact copy of an interface for this with `IReportHandler`. I this is needed for proper DI handling looking at something like `PassToHandlerReportDropbox` which needs to get all the `IReportHandler` implementations injected but it feels a bit like a technical necessity instead of explicit design? I can't quite think of a better way to handle this right now though. There could be one common base interface that both are extending to avoid the duplicate Dropoff definitions but that only works if we're sure the definitions are always going to be coupled between `IReportHandler` and `IReportDropbox`.

Elscrux (commented):
I think that's a great start to model the new concepts we have here! :)

Mutagen.Bethesda.Analyzers.Engine/Reporting/Drops/EditorIdEnricher.cs
Noggog: For sure.  I just like keeping things clean as i go.  We can always refer to past history to get the old logic for when we do the meta data PR

Mutagen.Bethesda.Analyzers.Engine/Reporting/Drops/IReportDropbox.cs
Noggog: Yeah, right now it's because the interfaces are handled differently by the DI.   IReportDropbox is supposed to be a decorator pipeline where there's one but it wraps the others in a conveyor belt.  But then it shifts to a "broadcast" style for IReportHandler[] where we want to get all of the handlers that exist and loop/report all of them.

Not sure how to have that transition from decorator pipeline to broadcast while using the same interface.   We can definitely evolve it as we go, as it's all internal mechanisms to the engine

PR #146, opened 2024-09-28
IsolatedRecordTestFixture

### LINKED ISSUE #145: Test Framework Revamp

Noggog:
Added a IsolatedRecordTestFixture to help write tests for IsolatedRecordAnalyzers

The idea is you provide the criteria for the triggering scenario, criteria for making it no longer trigger, and then what topic(s) should be triggered.   The system does the test, and eventually (hopefully) will test that the transition between the triggering scenario and non-triggering would provoke the engine to rerun (essentially testing the member accessor correctness)

I also cut up the monolith test classes into smaller ones organized by record type folder in preparation for other analyzer tests being put in the same folders/


Elscrux:
First had to understand that `MutagenModAutoData` actually creates `AssetLink` attributes including `existing` in their name in the mock file system but I think the system in general is really elegant and should be easy to use. I can't quite say how well more complicated analyzers adapt to this but it's probably fine.

Elscrux (approved):

PR #156, opened 2024-10-06
FieldsOfInterest

### LINKED ISSUE #150: Mechanism to mark interesting fields outside of topic reports
True, we want to know what relates to the issues so we can know when to rerun the analyzer proper.. but doing it only when we report an issue won't work well.   True, when an issue is reported, we can detect when those fields change and rerun... but we can't do the reverse.   How does an analyzer not reporting anything know when to rerun to potentially report something?   Well.. it can't.  It'd have to run every time
We'll probably want to have some other function/area of the analyzers that specify what fields are interesting, rather than at the time of topic reporting

Noggog:
Adds an abstract function FieldsOfInterest to specify what fields the analyzer needs to be rerun for


Noggog:
Also added in the finishing touches of #148

Mutagen.Bethesda.Analyzers.Skyrim/Record/Activator/MineOreAnalyzer.cs
```diff
@@ -85,7 +80,13 @@ public void AnalyzeRecord(ContextualRecordAnalyzerParams<IActivatorGetter> param
...
         param.AddTopic(
-            IncorrectVeinOre.Format(ore),
-            x => x.VirtualMachineAdapter);
+            IncorrectVeinOre.Format(ore));
+    }
+    public IEnumerable<Func<IActivatorGetter, object?>> FieldsOfInterest()
+    {
+        yield return x => x.EditorID;
+        yield return x => x.VirtualMachineAdapter!.Scripts;
+        yield return x => x.EditorID;
```
Elscrux: EditorID is used as field of interest multiple times

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/RaceAddonsAnalyzer.cs
```diff
@@ -59,8 +59,16 @@ public void AnalyzeRecord(ContextualRecordAnalyzerParams<IArmorGetter> param)
...
         }
     }
+
+    public IEnumerable<Func<IArmorGetter, object?>> FieldsOfInterest()
+    {
+        yield return x => x.TemplateArmor;
+        yield return x => x.MajorFlags;
+        yield return x => x.Race;
+        yield return x => x.Armature;
+        yield return x => x.Race;
```
Elscrux: Race is used as field of interest twice

Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
```diff
@@ -27,17 +27,20 @@ public void AnalyzeRecord(IsolatedRecordAnalyzerParams<ISkyrimMajorRecordGetter>
...
-                        x => x);
+                        InvalidConditionReference.Format(conditionData.Function.ToString()));
                     break;
             }
         }
     }
+
+    public IEnumerable<Func<ISkyrimMajorRecordGetter, object?>> FieldsOfInterest()
+    {
+        yield break;
```
Elscrux, line 42: We'll have to get back to this later when we have a good way to deal with conditions

Mutagen.Bethesda.Analyzers.Skyrim/Record/Constructible/MissingCreatedObjectAnalyzer.cs
```diff
@@ -18,8 +18,11 @@ public void AnalyzeRecord(IsolatedRecordAnalyzerParams<IConstructibleObjectGette
         if (param.Record.CreatedObject.IsNull)
         {
             param.AddTopic(
-                formattedTopicDefinition: MissingCreatedObject.Format(),
-                memberExpression: x => x.CreatedObject);
+                formattedTopicDefinition: MissingCreatedObject.Format());
```
Elscrux: I think we can get rid of this specific parameter mention

Mutagen.Bethesda.Analyzers.Skyrim/Record/Constructible/MissingEnchantedPerkAnalyzer.cs
```diff
@@ -45,7 +45,11 @@ public void AnalyzeRecord(IsolatedRecordAnalyzerParams<IConstructibleObjectGette
         }
 
         param.AddTopic(
-            formattedTopicDefinition: MissingCreatedObject.Format(),
-            memberExpression: x => x.Conditions);
+            formattedTopicDefinition: MissingCreatedObject.Format());
```
Elscrux: I think we can get rid of this specific parameter mention

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/MerchantAnalyzer.cs
```diff
@@ -29,7 +29,12 @@ editorId is not null
...
         param.AddTopic(
-            MerchantWithoutSpecialization.Format(),
-            x => x.Factions);
+            MerchantWithoutSpecialization.Format());
+    }
+
+    public IEnumerable<Func<INpcGetter, object?>> FieldsOfInterest()
+    {
+        yield return x => x.Factions;
+        yield return x => x.Factions;
```
Elscrux: Factions are used twice

Elscrux (approved):
The changes on the topic report syntax will make it a lot less cumbersome to write analyzers, that's great! I also understand why you added fields of interest per analyzer, not per topic. In the end the whole analyzer will need to run again based on these fields and that is what matters in this case, topics are a concept that's relevant elsewhere.

PR #159, opened 2024-10-12
Parallel

Noggog:
Added a lot of multithreading concepts

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Engine/Engines/ContextualEngine.cs
+++ b/Mutagen.Bethesda.Analyzers.Engine/Engines/ContextualEngine.cs
@@ -1,4 +1,4 @@
-﻿using Mutagen.Bethesda.Analyzers.Drivers;
+using Mutagen.Bethesda.Analyzers.Drivers;
 using Mutagen.Bethesda.Analyzers.SDK.Drops;
 using Mutagen.Bethesda.Environments.DI;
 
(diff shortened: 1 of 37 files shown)


Elscrux:
`CsvReportHandler` will need to sync access on the csv file

Mutagen.Bethesda.Analyzers.Engine/Engines/ContextualEngine.cs
```diff
@@ -58,16 +58,27 @@ public async Task Run()
...
             }
         }
 
-        foreach (var driver in ContextualModDrivers.Drivers)
+        var contextualParam = new ContextualDriverParams(
+            env.LinkCache,
+            env.LoadOrder,
+            ReportDropbox);
+
+        await Task.WhenAll(ContextualModDrivers.Drivers.Select(driver =>
```
Elscrux: Should this run next to the other drivers that take a mod above? Probably not necessary or detrimental as the degree of parallelism is going to be quite high anyway I assume?

Noggog: Yeah, we can probably improve this further.   Before i had things doing `Enqueue` without the `AndWait`, and in that mode, it didn't matter as the drivers above would quickly queue their work, and before it was done, it would get to this snippet and start off its work as well

However, that was causing problems in some areas (like environments getting disposed before their work was done).  I swapped back to `EnqueueAndWait`, as that's a lot more stable.  I was just trying to hyperoptimize since a new TaskCompletionSource needs to be made for each one.   Probably worth it for now, though

PR #164, opened 2024-11-12
Config Updates

Elscrux:


Elscrux:
Closes #142

Elscrux:
Closes #137

Elscrux:
Closes #157

Mutagen.Bethesda.Analyzers.Skyrim/Record/Cell/Interior/Settlement/NoMerchantChestLocRefTypeAnalyzer.cs
```diff
@@ -22,12 +22,18 @@ public void AnalyzeRecord(ContextualRecordAnalyzerParams<ICellGetter> param)
     {
         var cell = param.Record;
 
+        var merchantChests = param.LinkCache.PriorityOrder.WinningOverrides<IFactionGetter>()
+            .Where(faction => faction.Flags.HasFlag(Bethesda.Skyrim.Faction.FactionFlag.Vendor))
+            .Select(faction => faction.MerchantContainer.FormKey)
+            .ToList();
```
Noggog: `ToHashSet()` will be better for the Contains calls to follow

Mutagen.Bethesda.Analyzers.Engine/Mutagen.Bethesda.Analyzers.Engine.csproj
```diff
@@ -13,4 +13,9 @@
       <PackageReference Include="Mutagen.Bethesda.Autofac" />
     </ItemGroup>
 
+    <ItemGroup>
+      <Compile Remove="Config\AnalyzerConfigProvider.cs" />
```
Noggog: Probably dont want these lines?

Mutagen.Bethesda.Analyzers.Engine/Config/ConfigDirectoryProvider.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Mutagen.Bethesda.Environments.DI;
+using Noggog;
+using Noggog.IO;
+
+namespace Mutagen.Bethesda.Analyzers.Config;
+
+public class ConfigDirectoryProvider(
+    IDataDirectoryProvider dataDirectoryProvider,
```
Noggog, line 8: Probably fine, but the reason you don't see me using this style more is that these fields are technically not readonly.  They're private to the class, but mutable, which is annoying.  Would be perfect API otherwise

Mutagen.Bethesda.Analyzers.Engine/Config/Topic/AnalyzerConfigReader.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    private readonly ConfigReader<ITopicConfig> _reader;
+    private readonly ILogger<TopicConfigReader> _logger;
+
+    public TopicConfigReader(
+        ConfigReader<ITopicConfig> reader,
+        ILogger<TopicConfigReader> logger)
+    {
+        _logger = logger;
+        _reader = reader
```
Noggog: This pattern of doing the registration in the ctor feels like a code smell to me.   

I think it would be better if we perhaps had all the classes that were going to add to the reader logic instead implement something like `IConfigReaderProcessorUnit<TConfig>` rather than being  IConfigReader<TConfig> themselves (aka, kill the decorator pattern aspect to it), and then did either:

# Factory Pattern
```cs
public class ConfigReaderFactory<TConfig>
{
   private readonly IConfigReaderProcessorUnit<TConfig>[] _readerPlugins;
   public ConfigReaderFactory(IConfigReaderProcessorUnit<TConfig>[] readerPlugins)
   {
       _readerPlugins = readerPlugins;
   } 

   public ConfigReader<TConfig> GetNewReader()
   {
       var ret = new ConfigReader<TConfig>();
       foreach (var plugin in _readerPlugins)
       {
           ret.Register(plugin);
       }
   }
}
```

# Self Mutation
```cs
public ConfigReader(
    IConfigReaderProcessorUnit<TConfig>[] readerPlugins,
    ILogger<ConfigReader<TConfig>> logger)
{
    _logger = logger;
       foreach (var plugin in readerPlugins)
       {
           this.Register(plugin);
       }
}
```

This feels a bit better as at least the reader is doing the mutation itself and then is "finalized", and has less external side effects

Mutagen.Bethesda.Analyzers.Engine/Config/Analyzer/AnalyzerConfig.cs
```diff
@@ -0,0 +1,37 @@
...
+    public FilePath? OutputFilePath { get; }
+}
+
+public interface IAnalyzerConfig : IAnalyzerConfigLookup
+{
+    void OverrideDataDirectory(DirectoryPath path);
+    void OverrideLoadOrderSetByDataDirectory(bool value);
+    void OverrideGameRelease(GameRelease release);
+    void OverrideOutputFilePath(FilePath filePath);
+    void OverrideLoadOrderSetToMods(IEnumerable<ModKey> mods);
```
Noggog, line 21: Mentioned this before, but I think we'll eventually want to kill this API for the built in C# record init style API.   The tests do rely on this current setup as they use mocks + NSubstitutes `Received` style API, but I think those would just need a refactor a bit.   Simple enough to do
```cs
[Theory]
[AnalyzerInlineData("environment.data_directory = C:/some/path")]
[AnalyzerInlineData(@"environment.data_directory = C:\some\path")]
public void TestDataDirectory(
    string line,
    IAnalyzerConfig config,
    AnalyzerConfigReader sut)
{
    sut.ReadInto(line.AsSpan(), config);
    config.Received(1).OverrideDataDirectory("C:/some/path");
}
```
to just do
```cs
[Theory]
[AnalyzerInlineData("environment.data_directory = C:/some/path")]
[AnalyzerInlineData(@"environment.data_directory = C:\some\path")]
public void TestDataDirectory(
    string line,
    AnalyzerConfig config,
    AnalyzerConfigReader sut)
{
    sut.ReadInto(line.AsSpan(), config);
    config.DataDirectoryPath
        .Should().Be(new DirectoryPath("C:/some/path"));
}
```
Which, id argue is better anyway to not rely on mocks (I went through a lonnng testing journey.. coming out the other side as a "mock minimally" guy)

-----------

Don't necessarily have to do these changes in this PR.  If you dont want, I can just crack open a card for later

Mutagen.Bethesda.Analyzers.Engine/Config/ConfigReader.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    private readonly ILogger<ConfigReader<TConfig>> _logger;
+    private readonly List<Func<TConfig, IReadOnlyList<string>, string, bool>> _processors = [];
+
+    public ConfigReader(ILogger<ConfigReader<TConfig>> logger)
+    {
+        _logger = logger;
+    }
+
+    public ConfigReader<TConfig> Register(Func<TConfig, IReadOnlyList<string>, string, bool> process)
```
Noggog: I think it'll be more functional if this just takes in a `Func<TConfig, IReadOnlyList<string>, string, bool>[]` in the ctor, rather than being mutable.

The idea being when the ConfigReader is made, it is being passed all the processors that will make it up right then and there, rather than always being open to random mutation at some point in the future.

This might be done either via the factory idea, or the labelling of `Func<TConfig, IReadOnlyList<string>, string, bool>` as a `IConfigReaderProcessorUnit<TConfig>` interface or whatever so that ConfigReader can inject `IConfigReaderProcessorUnit[]` as a dependency and get the list of subprocessors upfront that way

Mutagen.Bethesda.Analyzers.Tests/Config/AnalyzerConfigReaderTests.cs
```diff
@@ -8,33 +10,63 @@ namespace Mutagen.Bethesda.Analyzers.Tests.Config;
...
+    [AnalyzerInlineData("environment.data_directory = C:/some/path")]
+    [AnalyzerInlineData(@"environment.data_directory = C:\some\path")]
+    public void TestDataDirectory(
         string line,
         IAnalyzerConfig config,
         AnalyzerConfigReader sut)
     {
         sut.ReadInto(line.AsSpan(), config);
-        config.Received(1).Override(new TopicId("A", 123), Severity.Warning);
+        config.Received(1).OverrideDataDirectory("C:/some/path");
```
Noggog: If some of the other comments are done.. then this might need to change a bit.   If `AnalyzerConfigReader` is instead the `IConfigReaderProcessorUnit<AnalyzerConfig>`.. then it won't self inject into a reader anymore here.

Could refactor it a few ways, depending on how the API shakes out.. but there's always this trick, if needed:
```cs
public class InternalTestPayload
{
    public readonly IConfigReader<AnalyzerConfig> Reader;

    public InternalTestPayload(AnalyzerConfigProcessingUnit processor)
    {
         // Some common setup
         Reader = new ConfigReader<AnalyzerConfig>(new [] { processor });
    }
}

// ...

[Theory]
    [AnalyzerInlineData("environment.data_directory = C:/some/path")]
    [AnalyzerInlineData(@"environment.data_directory = C:\some\path")]
    public void TestDataDirectory(
        string line,
        AnalyzerConfig config,
        InternalTestPayload payload)
    {
        payload.Reader.ReadInto(line.AsSpan(), config);
        config.DataFolder.Should.Be(...);
```

Mutagen.Bethesda.Analyzers.SDK/Analyzers/IAnalyzer.cs
```diff
@@ -4,5 +4,6 @@ namespace Mutagen.Bethesda.Analyzers.SDK.Analyzers;
 
 public interface IAnalyzer
 {
+    AnalyzerId Id => new(GetType().FullName ?? throw new InvalidOperationException("Analyzer type has no FullName"));
```
Noggog, line 7: This being used for anything atm?

Noggog (commented):
Dumped a bunch of thoughts on this.   Might see if i can do some of the suggested tweaks myself, as the current code works well enough.. and just some architectural personal preference here and there.

Mutagen.Bethesda.Analyzers.Engine/Config/ConfigDirectoryProvider.cs
Elscrux: Yeah, primary constructors are a bit flawed. I like to use them anyway because I know about their problems but of course that can lead to problems. Let me know if you want to change it.

Mutagen.Bethesda.Analyzers.Engine/Config/ConfigReader.cs
Elscrux: > The idea being when the ConfigReader is made, it is being passed all the processors that will make it up right then and there, rather than always being open to random mutation at some point in the future.

My plan to go sort of have a Builder pattern for this as it creates a new copy of the ConfigReader every time you register new functions to prevent mutations of any objects that you have. But yeah, also not ideal.

Mutagen.Bethesda.Analyzers.SDK/Analyzers/IAnalyzer.cs
Elscrux: I think it would need to be used when we want analyzer specific settings. Didn't have any of those yet though I'm realizing. Can you think of any applicable settings?

Mutagen.Bethesda.Analyzers.Engine/Config/Topic/AnalyzerConfigReader.cs
Elscrux: I think the Factory code here would need to save the resulting reader object from each `Register()` call itself as it's returning a new object. So the reader is never mutated when you call `Register()`, it just creates an "extended" version of itself, adds the the passed function to the extended version's processing steps and returns the new object. The name "Register" isn't really telling that here though I think.

**Nevermind, I didn't realize that the <TConfig> for the factory pattern and self mutation would already ensure that we always have the right processor steps for the right config so no manual resolving via DI or similar would be needed.
I think I'll go with the self mutation approach!** 

I'm not against changing the design of this implementation but I did like that you could simply inject the specific type of reader that you want like `AnalyzerConfigReader` to get all what you need. The factory has the extra step of calling the right method and it would need to know which config reader should use which processor units so we would put a lot of logic in there to handle that right? Similarly the self mutation approach would need some other class that is capsuling this class with a method to dynamically resolve the processor units it wants to inject into each reader via DI. At least as far as I can see?

Capsuling each processing step in an object could be nice from the perspective that you're not just dealing with raw Funcs but that could probably also be added independently if we want that. Though of course it kind of inflates the codebase when there is a class each, and they would need to be resolved via DI so they can also use things like loggers. Not quite sure, seems like everything has it's upsides and downsides 😄
What do you think are the biggest problems with the ctor registrations in the current decorator pattern with builder-like registrations that at least work with immutable readers?

Mutagen.Bethesda.Analyzers.Engine/Config/Analyzer/AnalyzerConfig.cs
Elscrux: Yeah I also prefer that! Didn't want to change the established way tests were written in this project though haha. I can change it when we have figured out the rest here :)

Mutagen.Bethesda.Analyzers.SDK/Analyzers/IAnalyzer.cs
Noggog: I think we should focus on "topics" for the end user's API.  I think what analyzer is handling what topic and how is an implementation detail that shouldn't be exposed too much.    "Oh.. topic ABC relates to reporting bug X.   It used to be fished for and reported via analyzer G, but now H does it as it's already doing this other thing and it's super optimized to also check X while it's there."

We dont want users to know about G or H to where the refactor would cause problems, i think?

Mutagen.Bethesda.Analyzers.Engine/Config/Topic/AnalyzerConfigReader.cs
Noggog: > I think I'll go with the self mutation approach!

This is a similar pattern to how the analyzers are wired together, where `ContextualDriver` requests all the `IContextualAnalyzer[] contextualAnalyzers` from the container.

> What do you think are the biggest problems with the ctor registrations in the current decorator pattern with builder-like registrations that at least work with immutable readers?

My current reasons are:
- I like doing functional programming where I can.  Just a learned preferred practice.  Unexpected mutations introduce edge cases
- There's nothing stopping the system from instantiating a latecomer that does a registration after the fact, at which point the parent config reader's logic suddenly changes due to the registration that suddenly occurred.    Probably wont happen.. but requesting all the `IConfigReaderProcessorUnit` ahead of time is just much clearer.  We know all the pieces that will make up the reader at instantiation time, and here they all are.   It just closes off whole categories of bugs.
- Confusion on the decoration order.   Whos the top decoration?  Whos the nested decoration?  Does it matter?  Can it be strung up wrong if done in the wrong order?   Lots to think about.   Vs injecting in `IConfigReaderProcessorUnit[]` is more easy to think aobut.   We're making the config reader, and we have 8 plugins that it found to iterate and use right here.  Easy

Noggog (approved):

PR #166, opened 2024-11-23
Make data directory config source optional

Elscrux:
Useful in case the data directory doesn't exist like in a CI environment. A custom path for a data directory can still be passed via config entry. It's just a workaround so potentially we might want to create a better solution for this problem.

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Engine/Config/ConfigDirectoryProvider.cs
+++ b/Mutagen.Bethesda.Analyzers.Engine/Config/ConfigDirectoryProvider.cs
@@ -21,10 +21,13 @@ public class ConfigDirectoryProvider(
             }
             catch (Exception)
             {
-                yield break;
+                dataDirectory = null;
             }
 
-            yield return dataDirectory.Value;
+            if (dataDirectory.HasValue)
+            {
+                yield return dataDirectory.Value;
+            }
         }
     }
 }
--- a/Mutagen.Bethesda.Analyzers.Engine/Config/ConfigDirectoryProvider.cs
+++ b/Mutagen.Bethesda.Analyzers.Engine/Config/ConfigDirectoryProvider.cs
@@ -13,7 +13,18 @@ public class ConfigDirectoryProvider(
         get
         {
             yield return currentDirectoryProvider.CurrentDirectory;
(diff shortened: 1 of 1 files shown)


Mutagen.Bethesda.Analyzers.Engine/Config/ConfigDirectoryProvider.cs
```diff
@@ -13,7 +13,18 @@ public IEnumerable<DirectoryPath> ConfigDirectories
...
-            yield return dataDirectoryProvider.Path;
+
+            DirectoryPath? dataDirectory;
+            try
+            {
+                dataDirectory = dataDirectoryProvider.Path;
+            }
+            catch (Exception)
+            {
+                yield break;
```
Noggog: Let's swap this to `dataDirectory = null;`
and then follow up with a null check before the yield return.

This will be less prone to errors in future changes when, say, someone slaps in a 3rd source.  If they dont notice this `yield break`, the 3rd source might not be considered if the data dir is not present

Noggog (changes requested):
Can you walk me through when the data directory provider would be throwing?  Would it be when the mutagen provided game environment cannot find the game installation?

Elscrux:
> Can you walk me through when the data directory provider would be throwing? Would it be when the mutagen provided game environment cannot find the game installation?

Yeah, exactly. When the data directory is not present it would throw when you try to access `dataDirectoryProvider.Path`.

Mutagen.Bethesda.Analyzers.Engine/Config/ConfigDirectoryProvider.cs
Elscrux: Good idea

Noggog:
> Yeah, exactly. When the data directory is not present it would throw when you try to access dataDirectoryProvider.Path.

Hmm, not ideal, but not sure the best way to do it with the current API.   Ideally I'd like to have something that had a nullable data folder to indicate when it wasn't found, rather than a throw.   But the current Mutagen API is written as non-nullable, so best we can do for now without going down a rabbit hole

Noggog (approved):

PR #168, opened 2024-11-23
Remove game release config option

Elscrux:
Not needed because the GameRelease is always set via start argument anyway


Noggog (commented):
I think this change is fine enough for what the PR says it's trying to do.  However, I'm still uncertain powers/purpose the config is trying to fulfill to know if it's a desirable change or not.

Ofc, a config is specifying lots of topic configuration overrides silent/error/warning, etc.   But how much of the command parameters is it trying to handle as well?  Should the config be in the business of setting GameEnvironment variables like Release/Data Directory/ Load order at all?   If so, should it not go all the way and be able to fully define everything that would drive an environment, such as GameRelease?

And if it's able to define GameRelease along with all the other game environment details.. perhaps one of the modes of the `RunAnalyzersCommand` shouldn't require GameRelease to be defined as a parameter at all.   This style would have a parameter pointing to a config explicitly, or omit it entirely and locate it by "current directory" presence?

In that case, this PR should be reversed, so that the config has game release again, and the command is what is changed to have GameRelease more optional?

------------

Just to throw one edge case in there that came to my head, how do we want to handle the following situation in the current setup:
- Command line does not provide data directory
- Game installation is not present, so it cannot be found implicitly
- Configuration in the current directory lists a data directory
- Does a config in the data directory get picked up and loaded in?  I think it would get skipped over

------ 

Whatever we decide, i think it would be good to write out some tests as a way of document our intentions (as well as showing they work as intended)

Elscrux:
Very good questions! I also encountered this issue when I built the extended config system. I think it would be nice to be able to specify all environment related aspects in a config tbh, just because arguments make it very cumbersome and longwinded. Of course a config needs more initial setup than just adding some arguments when calling the executable but I think we can accept that? Of course that's assuming that the config would totally replace some of the setting that you can currently also pass via arguments. Do you think we should allow both or just one?

> Just to throw one edge case in there that came to my head, how do we want to handle the following situation in the current setup:
>    Command line does not provide data directory
    Game installation is not present, so it cannot be found implicitly
    Configuration in the current directory lists a data directory
    Does a config in the data directory get picked up and loaded in? I think it would get skipped over

Yeah, I think we would skip  the one in the data directory.
In case we wouldn't have arguments to set the environment and just the config this would also make the PR #166 unnecessary as we wouldn't look for a config file in the data directory. That would also remove this edge case but of course forces everyone to use config files and an analyzerconfig in the in the current directory.

Noggog:
Is this obsolete with the #170 PR?

Elscrux:
Yeah it's not needed anymore

PR #170, opened 2025-01-10
Redesign Run Config

Elscrux:
- Run config (previously analyzer config) will now be passed via argument --RunConfigPath
- Replace LoadOrderSetByDataDirectory in favor of LoadOrderSetToMods
- Additionally filter analyzers passed to drivers in `IAnalyzerProvider` to filter any analyzers that don't have any enabled topics


Mutagen.Bethesda.Analyzers.Testing/TestAnalyzer.cs
```diff
@@ -16,7 +18,10 @@ public void Analyze(ContextualAnalyzerParams param)
     {
         foreach (var topic in Topics)
         {
-            param.AddTopic(topic.Format());
+            param.AddTopic(
+                ModKey.Null,
+                new MajorRecordIdentifier { FormKey = FormKey.Null },
```
Noggog, line 23: I think we can re-add the `ContextualAnalyzerParams.AddTopic` alternative without the ModKey/Record parameters, so that calls like this don't need to "fake it"?

I guess that'd make it a little easier for contextual analyzer devs to avoid passing in the desired information

Mutagen.Bethesda.Analyzers.SDK/Analyzers/IsolatedRecordFrameAnalyzerParams.cs
```diff
@@ -34,8 +39,20 @@ public void AddTopic(
         IFormattedTopicDefinition formattedTopicDefinition,
         params (string Name, object Value)[] metaData)
     {
+        string? editorId = null;
+        if (Frame.TryFindSubrecord(RecordTypes.EDID, out var edid))
+        {
+            editorId = edid.AsString(GameConstants.Get(_mod.GameRelease).Encodings.NonLocalized);
+        }
```
Noggog: I think we should move this logic upstream to whoever is driving.   The RecordFrameDriver making IsolatedRecordFrameAnalyzerParams knows the mod, the frame, and all that.  It can do the EDID extraction and MajorRecordIdentifier construction once ahead of time and just pass it in?    Feels more appropriate, rather than doing the work for each AddTopic

Mutagen.Bethesda.Analyzers.Engine/Config/Run/RunConfigProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Mutagen.Bethesda.Analyzers.Config.Run;
+
+public class RunConfigProvider
```
Noggog: Seems unused at the moment?  Or am i missing where it's being utilized?

Noggog (approved):
Looking good!   Added some spruce ideas, but none fundamental blockers

Mutagen.Bethesda.Analyzers.Testing/TestAnalyzer.cs
Elscrux: The idea was to make it mandatory to pass in all the information. Otherwise it might be easy to forget it and to just use the other variant where you don't have to pass this info.
For the test I think it might be fine to fake it, or have a test-only extension method that allows this workaround?

Mutagen.Bethesda.Analyzers.Engine/Config/Run/RunConfigProvider.cs
Elscrux: I think it can be removed the way the run config is used now.

Elscrux:
> Looking good! Added some spruce ideas, but none fundamental blockers

Thanks for the review! All should be fixed now :)

PR #173, opened 2025-02-05
Blacklisted mods provider

Elscrux:
Adds `IBlacklistedModsProvider` to exclude mods that shouldn't be analyzed.

For `ContextualAnalyzer` it just filters outgoing reports, as they might use records from these mods in their own analyze logic.

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Cli/Modules/AnalyzerCommandModule.cs
+++ b/Mutagen.Bethesda.Analyzers.Cli/Modules/AnalyzerCommandModule.cs
@@ -8,6 +8,7 @@ using Mutagen.Bethesda.Analyzers.Cli.Args;
 using Mutagen.Bethesda.Analyzers.Cli.Overrides;
 using Mutagen.Bethesda.Analyzers.Config;
 using Mutagen.Bethesda.Analyzers.Config.Run;
+using Mutagen.Bethesda.Analyzers.Drivers;
 using Mutagen.Bethesda.Analyzers.Reporting.Handlers;
 using Mutagen.Bethesda.Environments.DI;
 using Mutagen.Bethesda.Plugins;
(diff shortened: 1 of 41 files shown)


Mutagen.Bethesda.Analyzers.Cli/Modules/AnalyzerCommandModule.cs
```diff
@@ -72,5 +72,14 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<NullPluginListingsPathProvider>().As<IPluginListingsPathProvider>();
             builder.RegisterType<NullCreationClubListingsPathProvider>().As<ICreationClubListingsPathProvider>();
         }
+
+        if (command.BlacklistedMods is not null)
```
Noggog, line 76: Does this need an `else` where it registers `InjectedBlacklistedModsProvider` with no mods?  I think `FilterBlacklistedReports` will fail to resolve otherwise.

Alternatively, we could just not register `FilterBlacklistedReports` if there's nothing blacklisted.  But i'd have to look closer at the wiring to see if that'd be annoying to do

Noggog (approved):
One thing to clarify a bit is that this is blacklisting the mods from the "LO".  So engines that loop all mods and run some analyzers will skip the blacklisted mods.

Does it also skip overrides of blacklisted mods in non-blacklisted mods?
If `ModA.esp` overrides a record from `BlacklistedMod.esp`, do engines skip just that record when processing `ModA.esp`?

I imagine not.   But might be good to clarify the terms a bit.  Maybe `BlacklistedFromLoadOrderMods`?  Dunno

Elscrux:
> One thing to clarify a bit is that this is blacklisting the mods from the "LO". So engines that loop all mods and run some analyzers will skip the blacklisted mods.
> 
> Does it also skip overrides of blacklisted mods in non-blacklisted mods? If `ModA.esp` overrides a record from `BlacklistedMod.esp`, do engines skip just that record when processing `ModA.esp`?
> 
> I imagine not. But might be good to clarify the terms a bit. Maybe `BlacklistedFromLoadOrderMods`? Dunno

I think it is fair to keep analyzing this record in that case. At least for my use case the scope of what the mod does in general is what I'm after, not that it specifically just checks for the new records. I think that also makes sense as a lot of mods mostly just touch existing mods. So if you would like to analyze "MyAwesomeLeveledListOverhaul.esp" and only that, nothing else, then you couldn't do that properly as it basically just touched existing records.

Mutagen.Bethesda.Analyzers.Cli/Modules/AnalyzerCommandModule.cs
Elscrux: It actually shouldn't. It's a implicit so maybe I should add a comment, but the logic in `RunConfigModule` which runs before this will always register something for `IBlacklistedModsProvider`, so we're always safe there.
And if you had an entry in the config file, you don't want to overwrite that in the command logic where you might not find a setting for this where you would then set your blacklisted mods to an empty list again.

Noggog:
Sounds good.   No changes then?

Elscrux:
> Sounds good. No changes then?

Yeah, I think we're good to go

PR #188, opened 2025-05-03
Analyzer updates

Elscrux:
Latest fixes and additions for analyzers that I noticed in the last bit. Best to check out per commit.

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim/Contextual/UnownedBedAnalyzer.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/Contextual/UnownedBedAnalyzer.cs
@@ -24,6 +24,8 @@ public class UnownedBedAnalyzer : IContextualAnalyzer
 
             foreach (var placedObject in cell.GetAllPlaced(param.LinkCache).OfType<IPlacedObjectGetter>())
             {
+                if (placedObject.IsDeleted) continue;
+
                 // Owned beds are not a problem
                 if (!placedObject.Owner.IsNull) continue;
 
--- a/Mutagen.Bethesda.Analyzers.Skyrim/Contextual/UnownedBedAnalyzer.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/Contextual/UnownedBedAnalyzer.cs
@@ -30,7 +30,7 @@ public class UnownedBedAnalyzer : IContextualAnalyzer
                 if (!param.LinkCache.TryResolve<IFurnitureGetter>(placedObject.Base.FormKey, out var furniture)) continue;
                 if (!furniture.IsBed()) continue;
 
-                var context = param.LinkCache.ResolveSimpleContext(placedObject);
+                var context = param.LinkCache.ResolveSimpleContext<IPlacedObjectGetter>(placedObject.FormKey);
                 param.AddTopic(
                     context.ModKey,
                     placedObject,
--- a/Mutagen.Bethesda.Analyzers.Skyrim/Record/Key/WeightValueAnalyzer.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/Record/Key/WeightValueAnalyzer.cs
@@ -21,6 +21,7 @@ public class WeightValueAnalyzer : IIsolatedRecordAnalyzer<IKeyGetter>
     public void AnalyzeRecord(IsolatedRecordAnalyzerParams<IKeyGetter> param)
     {
         var key = param.Record;
+        if (key.IsDeleted) return;
 
         if (key is not { Weight: 0 })
         {
--- a/Mutagen.Bethesda.Analyzers.Skyrim/Util/InvalidCharactersAnalyzerUtil.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/Util/InvalidCharactersAnalyzerUtil.cs
@@ -1,4 +1,8 @@
-﻿namespace Mutagen.Bethesda.Analyzers.Skyrim.Util;
+﻿using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Plugins.Records;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Util;
 
 public static class InvalidCharactersAnalyzerUtil
 {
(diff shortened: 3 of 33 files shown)


Noggog:
Tangent, but I wouldn't mind seeing some unit tests being added here and there for some of these.   Stuff like the `Only report actual duplicate footsteps, not existing ones` fix could/should have some tests showing/flexing the new behavior works as expected, etc.     [For the isolated ones, we got a decent text fixture set up so writing tests shouldn't be a huge time sink](https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/blob/dev/Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/Quests/AliasIDAnalyzerTest.cs)

Would get good to get a similar setup going for contextual as well so we can start filling those out?   Always a bit of a pain writing tests, but I think they'll go a long way for the resiliency of the project longer term

Mutagen.Bethesda.Analyzers.Skyrim/Contextual/UnownedBedAnalyzer.cs
```diff
@@ -24,13 +24,15 @@ public void Analyze(ContextualAnalyzerParams param)
...
+                if (placedObject.IsDeleted) continue;
+
                 // Owned beds are not a problem
                 if (!placedObject.Owner.IsNull) continue;
 
                 if (!param.LinkCache.TryResolve<IFurnitureGetter>(placedObject.Base.FormKey, out var furniture)) continue;
                 if (!furniture.IsBed()) continue;
 
-                var context = param.LinkCache.ResolveSimpleContext(placedObject);
+                var context = param.LinkCache.ResolveSimpleContext<IPlacedObjectGetter>(placedObject.FormKey);
```
Noggog, line 35: Hmm, I can probably add another overload in Mutagen to avoid needing to do this.  Should be smart enough to identify that `placedObject` is `IPlacedObjectGetter` and use that as the type

Mutagen.Bethesda.Analyzers.Skyrim/Record/Key/WeightValueAnalyzer.cs
```diff
@@ -21,6 +21,7 @@ public class WeightValueAnalyzer : IIsolatedRecordAnalyzer<IKeyGetter>
     public void AnalyzeRecord(IsolatedRecordAnalyzerParams<IKeyGetter> param)
     {
         var key = param.Record;
+        if (key.IsDeleted) return;
```
Noggog, line 24: Hmm, do we perhaps want to consider not running isolated analyzers on deleted records?  Contextual would ofc need to handle themselves, still.

Isolated record analyzers shouldn't really be run on a deleted record,, I don't think?  We'd maybe make one specialized analyzer to check if any deleted record had context which would be a warning (it shouldnt have any)

Mutagen.Bethesda.Analyzers.Skyrim/Util/InvalidCharactersAnalyzerUtil.cs
```diff
@@ -15,4 +19,16 @@ public static class InvalidCharactersAnalyzerUtil
     public static Dictionary<string, string> InvalidStringsOneLiner { get; } = InvalidStrings
         .Concat(new Dictionary<string, string> { { "\r", "" }, { "\n", "" }, { "  ", " " } })
         .ToDictionary();
+
+    public static char[] InconsistentStrings { get; } = ['[', ']'];
+
+    public static void CheckInconsistentCharacters<T>(IsolatedRecordAnalyzerParams<T> param, string text, TopicDefinition<string> topic) where T : IMajorRecordGetter
```
Noggog, line 25: Maybe beating a dead horse, but I think we should hammer out unit tests all around, but especially for things like this.

It's more than just confirming correctness, I think.  In an open source project like this where communication over time and space is often hard, unit tests provide almost "reasoning and expectation documentation"

For ex, let's say someone wants to refactor this component in 4 years.  They can of course see what it's currently doing.  

But what they can't see is YOUR intention.  They can't see what it is supposed to do.  They can only see what it is currently doing.   It's hard to sus out what is coincidentally in the code and can be changed during renovations vs what was intentionally in the code and should remain.    Is X actually something important? Or just a coincidental byproduct of how it's written currently?

For ex, someone might have trouble figuring out why `{ "  ", " " }` is in the InvalidStringsOneLiner (I certainly need to stare at it a bit to know what it's trying to accomplish).

By having unit tests, it more clearly lays out your intention for the code.   By making a unit test flexing X, you're explicitly stating "this is something important for the code to do that I intentionally set it up to do".  By having it, you're not only are confirming correctness, you're also making it easier on future maintainers to see inside your brain and know the bounds they can refactor things safely within

Noggog (approved):

Mutagen.Bethesda.Analyzers.Skyrim/Record/Key/WeightValueAnalyzer.cs
Elscrux: I thought about that too, but then wasn't sure if there might be some analyzers that still want to do it for some reason. But I think I'd welcome the change if we can have a variant that can also check deleted records if necessary. I always forgot this check and it was annoying :D

Mutagen.Bethesda.Analyzers.Skyrim/Contextual/UnownedBedAnalyzer.cs
Elscrux: I've got a lot of bad experienced with not adding explicit getter types into resolve calls because they end up trying to look for the setter variants (which don't exist in the binary overlay) too many times. Might be that I'm getting too paranoid with this, just has been my experience to be better safe than sorry with this 😅

Noggog: Yeah, for sure.   Adding it explicitly is never bad if you're used to adding it

Mutagen.Bethesda.Analyzers.Skyrim/Util/InvalidCharactersAnalyzerUtil.cs
Elscrux: Will definitely need to do more of that when we have all the test setup ready 💯

Noggog:
Merging this in for now?

Elscrux:
> Merging this in for now?

Probably, yeah. I can do the tests in another branch.

PR #495, opened 2025-08-31
Lots more analyzers + analyzer api

Elscrux:
Sorry for this beast of a PR, I don't want to bug you with too frequent ones but then in the end I also don't want to create lots of PRs so I end up with this :D  Lmk if you prefer a different format.

I got discussions set up for all the analyzers added here. No unit tests sadly (would still need the contextual analyzer testing framework figured out for a lot of those).

The most relevant part here is the last commit though - which is the reason I'm committing this. I want to create an api that we can publish as a nuget package so we can call the analyzer suite externally. That'd be a huge win also for visibility!

Lmk your thoughts on the current api going via the game environment, and everything else. The current setup was a bit of a bit to figure out the proper DI for this but it seems to be working relatively well as it is rn.


Noggog:
Always happy to get PRs!  Big or small.  Will be looking at it

Mutagen.Bethesda.Analyzers.Api/Mutagen.Bethesda.Analyzers.Api.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
Noggog: I think we can just make the project itself `Mutagen.Bethesda.Analyzers.csproj`?  No need for the `.Api` suffix, imo.

Mutagen.Bethesda.Analyzers.Skyrim/Util/VoiceTypeAssetLookupUtil.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Util;
+
+public static class VoiceTypeAssetLookupUtil
+{
+    private static readonly Lock Lock = new();
+    private static VoiceTypeAssetLookup? _voiceTypeAssetLookup;
+
+    public static VoiceTypeAssetLookup GetVoiceTypeAssetLookup(ILinkCache linkCache)
+    {
+        // TODO: Replace with something that is updated when the link cache is updated
```
Noggog, line 14: Can you talk me through what the concern is that requires a lock?
Is it because it might be a mutable cache passed in?   
I've been coming at analyzers from the angle that it'll be mostly all functional patterns (eg, immutable)

Mutagen.Bethesda.Analyzers.Skyrim/Util/VoiceTypeAssetLookupUtil.cs
```diff
@@ -0,0 +1,28 @@
...
+        lock (Lock)
+        {
+            if (_voiceTypeAssetLookup is null)
+            {
+                var immutableAssetLinkCache = linkCache.CreateImmutableAssetLinkCache();
```
Noggog, line 19: I think the ContextualRecordAnalyzerParams and friends can expose the asset link cache, rather than this making it?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Responses/LinksToDifferentSpeakerAnalyzer.cs
```diff
@@ -0,0 +1,60 @@
...
+            Severity.Error)
+        .WithFormatting<IDialogTopicGetter>("Topic has response that links to topic {0} that has no speakers in common with the responses linking to it");
+
+    public IEnumerable<TopicDefinition> Topics { get; } = [LinksToDifferentSpeaker];
+
+    public void AnalyzeRecord(ContextualRecordAnalyzerParams<IDialogTopicGetter> param)
+    {
+        var dialogTopic = param.Record;
+
+        var voiceTypeAssetLookup = VoiceTypeAssetLookupUtil.GetVoiceTypeAssetLookup(param.LinkCache);
```
Noggog, line 23: I wonder if we can add some "downstream" cache concept.   For ex, the params could provide something like
`param.RetrieveCache<VoiceTypeAssetLookup>()`
where `VoiceTypeAssetLookup` implements an interface like
```
public interface IDownstreamCache
{
     void Prep(ILinkCache linkCache, IAssetLinkCache assetCache);
}
```

The params object could then be the one to do/handle most of the work in VoiceTypeAssetLookupUtil.  Additionally, the engine would be better able to invalidate it all when the linkcache changed

Noggog (commented):
Some thoughts out.  More to look through

Mutagen.Bethesda.Analyzers.Api/Mutagen.Bethesda.Analyzers.Api.csproj
Elscrux: Sure! I thought the same for the nuget package (and added the property so it would be called that way), but i wasn't sure about the project itself.

Mutagen.Bethesda.Analyzers.Skyrim/Util/VoiceTypeAssetLookupUtil.cs
Elscrux: It has a lock because we are running analyzers in parallel and we might try to create multiple immutable caches at the same time when it is still null atm. Very wip type though and in the future this should probably be refactored with more thought put into it.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Responses/LinksToDifferentSpeakerAnalyzer.cs
Elscrux: Would that be an analyzers concept or a mutagen concept?
In general, yeah I think that would be something we should do! The current method is really not any good :D
This would handle both usage cache and voice type cache uses as they are both immutable and a snapshot of the current state.

Additionally, this is something the api should also capture in the future.
Here we need to think about if we assume that the link cache can update between every api call, or if we want to require users to specifically tell us when the link cache has updated. Or if we can somehow know ourselves, which would probably be best but currently that's not possible I'm sure.

Noggog: It would be a feature of the analyzer engine, and exposed via the parameter objects.  This would be similar to a container-like api, where you might do `param.ResolveCache<VoiceTypeAssetLookup>()`.   The analyzer would get one back without knowing how it was made, or knowing whether it was the same one as the last time it ran.

This would let the engine:
- Handle its construction.  We might make a component that `ICanMake<VoiceTypeAssetLookup>` that might itself `ResolveCache<IAssetLinkCache>()`, for ex
- Handle when it should be invalidated.  Engine could track when a new linkcache is required, and thus which downstream caches to purge so new ones are made

Mutagen.Bethesda.Analyzers.Skyrim/Util/VoiceTypeAssetLookupUtil.cs
Noggog: I think that container-esque system mentioned below would be the one to lock things internally as appropriate when fulfilling the requests

Mutagen.Bethesda.Analyzers.Skyrim/Util/UsageCacheUtil.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Mutagen.Bethesda.Plugins.Cache;
+using Mutagen.Bethesda.Plugins.Cache.Internals.Implementations;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Util;
+
+public static class UsageCacheUtil
```
Noggog, line 6: Fold this into the cache system we're brewing

Mutagen.Bethesda.Analyzers.Skyrim/Util/InvalidCharactersAnalyzerUtil.cs
```diff
@@ -7,6 +7,16 @@ namespace Mutagen.Bethesda.Analyzers.Skyrim.Util;
 
 public static class InvalidCharactersAnalyzerUtil
 {
+    public static Dictionary<char, string> InvalidStrings { get; } = new()
```
Noggog, line 10: Why is this a Dictionary?  All the usages seem to use it like a set

Mutagen.Bethesda.Analyzers.Skyrim/Util/TextReplacementUtil.cs
```diff
@@ -0,0 +1,70 @@
...
+using Mutagen.Bethesda.Strings;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Util;
+
+public static partial class TextReplacementUtil
+{
+    [GeneratedRegex(@"<(\w+)(?:.(\w+))?=(\w+)>", RegexOptions.IgnoreCase)]
+    public static partial Regex TextReplacementRegex { get; }
+
+    public static void CheckString(
```
Noggog, line 13: Not quite following what's happening here.  We're looping dialog text, and if it happens to contain `Global`, that means it's referencing global variables elsewhere?  What about alternative languages?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Relationship/RelationshipTypeAnalyzer.cs
```diff
@@ -0,0 +1,48 @@
...
+            "Courting with low relationship rank",
+            Severity.Suggestion)
+        .WithFormatting<Bethesda.Skyrim.Relationship.RankType>("Relationship type is Courting, but their rank {0} is lower than Acquaintance");
+
+    public IEnumerable<TopicDefinition> Topics { get; } = [CourtingRelationshipType];
+
+    public void AnalyzeRecord(ContextualRecordAnalyzerParams<IRelationshipGetter> param)
+    {
+        var relationship = param.Record;
+        if (relationship.IsDeleted) return;
```
Noggog: Wondering if we can just have the engine exclude deleted records entirely, rather than needing this everywhere?  Can we think of a reason we might not want that?   If so, maybe we flip it around so that analyzer interested in deleted records would need to flag itself somehow

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/PlacedObjectExtensions.cs
```diff
@@ -54,4 +64,88 @@ public static bool HasLocationRefType(this IPlacedObjectGetter placedObject, For
...
+                else
+                {
+                    // In case keyword or reference is not null, we check if it's a reference
+                    var placed = linkedRef.KeywordOrReference.TryResolve<IPlacedGetter>(linkCache);
+                    if (placed is not null)
+                    {
+                        return placed;
+                    }
+                }
+            }
```
Noggog, line 91: Is this just returning the first linked reference found that resolves?

I think it would be cleaner to split this up into two functions:
- ResolveIfLinksToReference where `IFormLinkGetter<IKeywordGetter> keyword` is required
- FindFirstResolvingLinkedReference which does this chunk

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/PlacedObjectExtensions.cs
```diff
@@ -54,4 +64,88 @@ public static bool HasLocationRefType(this IPlacedObjectGetter placedObject, For
...
+    }
+
+    public static ICellGetter? GetCell(this IPlacedGetter placed, ILinkCache linkCache)
+    {
+        if (!placed.ToLink().TryResolveSimpleContext(linkCache, out var context)) return null;
+        if (context.Parent?.Record is not ICellGetter cell) return null;
+
+        if (cell.IsInteriorCell()) return cell;
+
+        if (context.Parent?.Parent?.Record is not IWorldspaceGetter worldspace) return null;
```
Noggog: I suppose this works, since the parent nesting should be predictable
Can also swap to this
`if (!context.TryGetParent<IWorldspaceGetter>(out _)) return null;`
which will look the entire parent chain for the worldspace for you.   Same with the Cell retrieval above

Mutagen.Bethesda.Analyzers.Skyrim/Record/Placed/Object/LocRefTypeAnalyzer.cs
```diff
@@ -0,0 +1,294 @@
...
+    ];
+
+    public void AnalyzeRecord(ContextualRecordAnalyzerParams<IPlacedGetter> param)
+    {
+        var placed = param.Record;
+        if (placed.LocationRefTypes is null) return;
+
+        foreach (var locationRefType in placed.LocationRefTypes)
+        {
+            if (locationRefType.FormKey == FormKeys.SkyrimSE.Skyrim.LocationReferenceType.HouseBedRefType.FormKey)
```
Noggog: I think you can just use FormLink equality here (and lots of other places in this PR)
`if (locationRefType.Equals(FormKeys.SkyrimSE.Skyrim.LocationReferenceType.HouseBedRefType))`

No need to drill into the FormKey unless you like it that way.   If you use the `Equals`, it'll also protect against type mismatches.. which may or may not be desirable depending on the situation

Mutagen.Bethesda.Analyzers.Skyrim/Record/Placed/Npc/ScaleAnalyzer.cs
```diff
@@ -0,0 +1,41 @@
...
+        var placedNpc = param.Record;
+
+        if (placedNpc.IsDeleted) return;
+
+        var scaleNullable = placedNpc.Scale;
+        if (scaleNullable is null) return;
+
+        var scale = scaleNullable.Value;
+
+        if (Math.Abs(scale - 1) > 0.01)
```
Noggog: Can also consider the extension function EqualsWithin

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/HeadDataExtension.cs
```diff
@@ -0,0 +1,30 @@
...
+    public static bool IsNoseMorphAvailable(this IHeadDataGetter headData, uint noseIndex)
+    {
+        return headData.AvailableMorphs is { Nose: {} nose } && IsMorphAvailable(nose, noseIndex);
+    }
+
+    public static bool IsMorphAvailable(this IMorphGetter morph, uint index)
+    {
+        var div = (int)index / 8;
+        var mod = (byte)(index % 8);
+        var data = morph.Data[div];
```
Noggog, line 26: Is this something the Mutagen API could handle better?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/CarryPackageAnalyzer.cs
```diff
@@ -62,16 +82,81 @@ public void AnalyzeRecord(ContextualRecordAnalyzerParams<INpcGetter> param)
...
+                    param.AddTopic(
+                        NoCarryItemProperty.Format());
+                }
+            }
+        }
 
-        var carryItemPotionProperty = scriptEntry.GetProperty<IScriptObjectPropertyGetter>("CarryItemPotion");
-        if (carryItemPotionProperty is not null) return;
+        var placedNpc = UsageCacheUtil.GetUsageCache(param.LinkCache)
+            .GetUsagesOf<IPlacedNpcGetter>(npc).UsageLinks
```
Noggog, line 100: Why does this only look at the first PlacedNpc that references it?

Shouldnt it analyze all the PlacedNpcs for the criteria that raise topics?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/UniqueNpcPlacementAnalyzer.cs
```diff
@@ -0,0 +1,54 @@
...
+            "Unique NPC Placed Multiple Times",
+            Severity.Warning)
+        .WithoutFormatting("Unique NPC is placed multiple times in the world");
+
+    public IEnumerable<TopicDefinition> Topics { get; } = [PlacedNever, PlacedMultiple];
+
+    public void AnalyzeRecord(ContextualRecordAnalyzerParams<INpcGetter> param)
+    {
+        var npc = param.Record;
+        if (!npc.IsEligibleForTest(param.LinkCache)) return;
```
Noggog: I think this should be renamed to `IsUniqueActorType`

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/UniqueNpcPlacementAnalyzer.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Npc.Unique;
+
+public class UniqueNpcPlacementAnalyzer : IContextualRecordAnalyzer<INpcGetter>
+{
+    public static readonly TopicDefinition PlacedNever = MutagenTopicBuilder.FromDiscussion(
+            484,
+            "Unique NPC Never Placed",
+            Severity.Warning)
+        .WithoutFormatting("Unique NPC is never placed in the world");
```
Noggog: Feels like these are in the Suggestion realm?  Warnings being more something we know is malformed/wrong, but doesn't have identifiable ramifications, like list counters lying about the count.

This feels like a "hey, probably want to use this NPC, no?" slight nudge

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/UniqueNpcPlacementAnalyzer.cs
```diff
@@ -0,0 +1,54 @@
...
+
+    public static readonly TopicDefinition PlacedMultiple = MutagenTopicBuilder.FromDiscussion(
+            485,
+            "Unique NPC Placed Multiple Times",
+            Severity.Warning)
+        .WithoutFormatting("Unique NPC is placed multiple times in the world");
```
Noggog, line 21: Is this actually an issue?  What if they move around like some named dragon that haunts several locations?

Mutagen.Bethesda.Analyzers.Skyrim/Record/LeveledItem/ChanceNoneAnalyzer.cs
```diff
@@ -0,0 +1,47 @@
...
+
+public partial class ChanceNoneAnalyzer : IIsolatedRecordAnalyzer<ILeveledItemGetter>
+{
+    [GeneratedRegex(@"(\d+)$")]
+    public static partial Regex ChanceEditorIDRegex { get; }
+
+    public static readonly TopicDefinition<Percent, int> InvalidChanceNoneEditorID = MutagenTopicBuilder.FromDiscussion(
+            471,
+            "Invalid Chance in Editor ID",
+            Severity.Error)
```
Noggog: Is this just convention for readability?   I dont imagine it actually causes an Error in-game

Mutagen.Bethesda.Analyzers.Skyrim/Record/LeveledItem/ChanceNoneAnalyzer.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System.Text.RegularExpressions;
+using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Skyrim;
+using Noggog;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.LeveledItem;
+
+public partial class ChanceNoneAnalyzer : IIsolatedRecordAnalyzer<ILeveledItemGetter>
```
Noggog, line 9: Rename to ChanceNoneMatchesEditorIdAnalyzer?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/NoCombatStyleAnalyzer.cs
```diff
@@ -0,0 +1,34 @@
...
+using Mutagen.Bethesda.Skyrim;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Npc.Unique;
+
+public class NoCombatStyleAnalyzer : IContextualRecordAnalyzer<INpcGetter>
+{
+    public static readonly TopicDefinition NoCombatStyleFaction = MutagenTopicBuilder.FromDiscussion(
+            404,
+            "Npc has no Combat Style",
+            Severity.None)
```
Noggog, line 12: I dont think we should actually be using the None Severity.  I think that's mainly for filtering phases that modify severity to None to hide things

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/MultipleBedOwnershipAnalyzer.cs
```diff
@@ -0,0 +1,68 @@
...
+
+public class MultipleBedOwnershipAnalyzer : IContextualRecordAnalyzer<INpcGetter>
+{
+    public static readonly TopicDefinition<int> NpcOwnsMultipleBeds = MutagenTopicBuilder.FromDiscussion(
+            493,
+            "Npc Owns Multiple Beds",
+            Severity.Suggestion)
+        .WithFormatting<int>("Npc owns {0} beds");
+
+    public static readonly TopicDefinition NpcOwnsNoBeds = MutagenTopicBuilder.FromDiscussion(
```
Noggog, line 18: Maybe im not familiar with the IsEligibleForTest, but is this actually an issue?  I imagine a lot of Npcs dont have beds.   Is the fact that they're marked "actor" mean they always should have one as they're more town-based Npcs?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Topic/GiveGoldAnalyzer.cs
```diff
@@ -0,0 +1,65 @@
...
+public partial class GiveGoldAnalyzer : IIsolatedRecordAnalyzer<IDialogTopicGetter>
+{
+    [GeneratedRegex(@"\((\d+|<\w+>) (gold)\)", RegexOptions.IgnoreCase)]
+    private static partial Regex GoldRegex { get; }
+
+    public static readonly TopicDefinition<Language, string> GoldSpelledWrongTopic = MutagenTopicBuilder.FromDiscussion(
+            465,
+            "Gold Spelled Wrong in Topic",
+            Severity.Suggestion)
+        .WithFormatting<Language, string>("Topic prompt in {0} spells giving gold as '{1}' instead of 'gold'");
```
Noggog, line 18: Just out of curiosity, what types of string are we fishing for here that are wrong?  Why this mis-spelling different than the multitudes of other words that could be spelled wrong?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Cell/EncounterZoneAnalyzer.cs
```diff
@@ -0,0 +1,41 @@
...
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Skyrim;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Cell;
+
+public class EncounterZoneAnalyzer : IContextualRecordAnalyzer<ICellGetter>
+{
+    public static readonly TopicDefinition<IFormLinkGetter<ILocationGetter>, IFormLinkNullableGetter<ILocationGetter>> EncounterZoneLocationMismatch = MutagenTopicBuilder.FromDiscussion(
+            457,
+            "Encounter Zone / Location Mismatch",
```
Noggog: `/` suggests "or" to me?     Maybe `<->` or something?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Cell/Interior/EnemyLevelMultiplierAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
+﻿using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Skyrim;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Cell.Interior;
+
+public class EnemyLevelMultiplierAnalyzer : IContextualRecordAnalyzer<ICellGetter>
```
Noggog, line 7: This is the sort of analyzer im a bit wary of.  Too subjective.  Why 30%?  How would they change this if they wanted 40%?  What if someone is designing something to be a specific situation?  

Probably fine to keep.  But I do think we'll want to figure out some way to extract out analyzers to other packages so we can keep the "core" ones mostly objective

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/P3FloatExtension.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Noggog;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Extensions;
+
+public static class P3FloatExtension
```
Noggog, line 5: Probably can be migrated up to CSharpExt sometime

Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/Cells/ShowSkyAnalyzerTests.cs
```diff
@@ -8,24 +9,24 @@ namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.IsolatedRecords.Cells;
...
-            },
-            prepForFix: static cell =>
-            {
-                cell.EditorID = "TestCell";
-                cell.Flags = Cell.Flag.IsInteriorCell | Cell.Flag.ShowSky;
-                cell.Regions = [FormKeys.SkyrimSE.Skyrim.Region.WeatherMountains];
-            },
-            ShowSkyAnalyzer.ShowSkyWithoutRegion);
-    }
+    // [Theory, MutagenModAutoData]
```
Noggog, line 12: What made this not work anymore?

Mutagen.Bethesda.Analyzers.Api/IAnalyzer.cs
```diff
@@ -0,0 +1,206 @@
...
+
+public static class AnalyzerExtensions
+{
+    public static IAnalyzer Analyzer(this IGameEnvironment gameEnvironment, AnalyzerOptions? options = null)
+    {
+        return new Analyzer(gameEnvironment, options);
+    }
+}
+
+public interface IAnalyzer
```
Noggog: I think we should rename this.  I view an analyzer as all the many analyzers you've been writing.

I think this component is more of an AnalyzerRunner?

Mutagen.Bethesda.Analyzers.Api/IAnalyzer.cs
```diff
@@ -0,0 +1,206 @@
...
+public class Analyzer : IAnalyzer
+{
+    private readonly IGameEnvironment _gameEnvironment;
+    private readonly IWorkDropoff _workDropoff;
+    private readonly IContainer _container;
+
+    public IDriverProvider<IContextualDriver> ContextualModDrivers { get; }
+    public IDriverProvider<IIsolatedDriver> IsolatedModDrivers { get; }
+
+    public Analyzer(IGameEnvironment gameEnvironment, AnalyzerOptions? options = null)
```
Noggog: I think the ctor itself should be adjusted a bit.
- Inject IFileSystem into the ctor, rather than assuming a real filesystem
- IGameEnvironment is too constricting.  It needs a few things GameEnvironment might provide, but it doesn't actually need all that.  From the game environment, it needs:
   - GameRelease
   - LoadOrder
   - LinkCache

The helper functions going from the GameEnvironment to an analyzer can get the appropriate things and pass them along, rather than fundamentally requiring a GameEnvironment

On that note, i think we can provide static Factory functions on this class to expose the alternatives conveniently.   One that takes in those parameters exactly, one that takes in a game environment, maybe some other flavors.

Mutagen.Bethesda.Analyzers.Api/IAnalyzer.cs
```diff
@@ -0,0 +1,206 @@
...
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Plugins.Meta;
+using Mutagen.Bethesda.Plugins.Records;
+using Noggog.WorkEngine;
+
+namespace Mutagen.Bethesda.Analyzers.Api;
+
+public static class AnalyzerExtensions
+{
+    public static IAnalyzer Analyzer(this IGameEnvironment gameEnvironment, AnalyzerOptions? options = null)
```
Noggog: I prefer the static factories on `Analyzer.Factory` rather than an extension method on GameEnvironment.  But I guess it's not too harmful

Mutagen.Bethesda.Analyzers.Api/AnalyzerOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Mutagen.Bethesda.Analyzers.Config.Topic;
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+
+namespace Mutagen.Bethesda.Analyzers.Api;
+
+public class AnalyzerOptions : IMinimumSeverityConfiguration
+{
+    public Dictionary<TopicId, Severity> SeverityOverrides { get; set; } = new();
```
Noggog: Id rather this specify a TopicConfig that the user can pass in.  That way, if we add functionality to TopicConfig to help easier specification of things, the user can use it, rather than just dealing with a naked dictionary.

Mutagen.Bethesda.Analyzers.Api/AnalyzerOptions.cs
```diff
@@ -0,0 +1,11 @@
...
+    public Severity MinimumSeverity { get; set; } = Severity.Suggestion;
+    public int? NumberOfThreads { get; set; } = null;
```
Noggog: The internal API actually takes in IObservable<int> for the number of threads, letting an external program adjust them on the fly (if the settings are adjusted, for ex).  

Exposing IObservable<int?> would be the fully customizable way to expose it.  But that's annoying for basic users who dont understand observables.

So maybe some Factory alternatives.  Or a builder.  Hmm

Mutagen.Bethesda.Analyzers.Api/IAnalyzer.cs
```diff
@@ -0,0 +1,206 @@
...
+                {
+                    AnalyzerType = analyzer.GetType()
+                });
+            }, CancellationToken.None);
+        }));
+
+        await Task.WhenAll(isolated, contextual);
+
+        return collectionReportHandler.Results;
+    }
```
Noggog: This feels like it's doing too much and is in the weeds on details it shouldnt be concerned with so directly.   We should evolve our engine so we can just call it here, rather than having what is essentially the "glue" class being so directly involved in how to actually execute the logic.

We could instead improve ContextualEngine to take in a single record if that's of interest, rather than having mostly duplicate logic from ContextualEngine copy pasted here

Noggog (changes requested):
Some random thoughts on analyzers as I pass by, as per usual, haha.

Added some thoughts on the API bits as well!

Mutagen.Bethesda.Analyzers.Api/AnalyzerOptions.cs
Elscrux: Could potentially expose it as the interface `INumWorkThreadsController` directly so they can still just use `NumWorkThreadsConstant` to pass in a number? Might be exposing too much though.
Factory/builder could also be a thing, though for now I think we can probably leave it? Seems like a niche usecase.

Elscrux: Actually update on this I'm working on a builder for `AnalyzerRunner` where this is easy to expose. Problem is just that CSharpExt doesn't have a `INumWorkThreadsController ` implementation that accepts an IObservable.

Mutagen.Bethesda.Analyzers.Api/IAnalyzer.cs
Elscrux: Was mainly adding this one for visibility in mind. `IGameEnvironment` is what people use most so it's probably easy to get started with it there. But not necessarily the best for sure.

Mutagen.Bethesda.Analyzers.Api/IAnalyzer.cs
Elscrux: Yeah for sure!

Elscrux: Problem is that something like `ByGenericTypeRecordContextualDriver` is specifically designed to work on a whole load order. So we probably need to have some more alternative options there too, or do you have a different design in mind
If you want to play around with this a little yourself, feel free to do that :)

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/HeadDataExtension.cs
Elscrux: I think so yeah. It's basically lots of bitwise flags, so an array of bools would be good to expose here I think.

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/PlacedObjectExtensions.cs
Elscrux: I think from an api perspective it might be easiest to just call `GetLinkedReference`? Reduces the brainload.

Elscrux: Or `GetLinkedReference` and `GetLinkedReferenceWithKeyword` if we want to split it up.

Mutagen.Bethesda.Analyzers.Skyrim/Util/InvalidCharactersAnalyzerUtil.cs
Elscrux: For future fixers (we'll need to work on that api at some point). To know what to replace stuff with.

Mutagen.Bethesda.Analyzers.Skyrim/Util/TextReplacementUtil.cs
Elscrux: Yeah see this: https://ck.uesp.net/wiki/Text_Replacement
Pretty sure Global is something that all localizations need to use like that too, it's a specific editor term

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/UniqueNpcPlacementAnalyzer.cs
Elscrux: Creates issues when you have something like a quest that's supposed to work with "the npc" and when you have not one but two placements that's a problem. Also using something like a condition checking if someone is dead might be that you killed the one instance and now the other one is also regarded dead. Never good.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Relationship/RelationshipTypeAnalyzer.cs
Elscrux: Yeah we talked about it before and I definitely agree. Just wasn't sure if you wanted me to add behavior like this or if you were going to do it or so.

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/PlacedObjectExtensions.cs
Elscrux: Yeah should always be the same. Idk in TryGetParent you have a loop and more checks so technically less performant. Do you have a preference? I don't really care.

Mutagen.Bethesda.Analyzers.Skyrim/Record/LeveledItem/ChanceNoneAnalyzer.cs
Elscrux: Changed to suggestion, was already set up like that in the GitHub discussions

Mutagen.Bethesda.Analyzers.Skyrim/Record/LeveledItem/ChanceNoneAnalyzer.cs
Elscrux: I thought I'd keep it vague if we want to add more topics related to ChanceNone at some point?

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/NoCombatStyleAnalyzer.cs
Elscrux: This is something that you don't really need to care about usually so I thought I set it to none for this. I think there is like one other case with this severity so far. But I thought in case somebody actually wants to make sure this is the case for their setup then that is something they can bump up in severity. But usually you don't need to check it out and suggestion would already be too much for most cases.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/MultipleBedOwnershipAnalyzer.cs
Elscrux: They're unique npcs, so people that live a life in the world. They should usually have a bed.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Topic/GiveGoldAnalyzer.cs
Elscrux: just something liek gold spelled as uppercase "Ok, I'll take it (Give 10 Gold)" is wrong and "Ok, I'll take it (Give 10 gold)" is wrong and is right. Just something that we noticed in Beyond Skyrim and seeing as it's a vanilla consistency thing (vanilla does it wrong 1 time) I thought I might just include it for everybody.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Cell/Interior/EnemyLevelMultiplierAnalyzer.cs
Elscrux: This is basically a reminder in case you don't have any npcs in your dungeon that are not at the exact same level as the player, but also some that are maybe a bit higher level and more challenging.

The 30% are totally arbitrary and I agree it would be nice if you'd have options for something like this in the future to set it to 40% or whatever you want. Could be some kind of extension of `TopicConfig`.

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/P3FloatExtension.cs
Elscrux: All extension functions more or less should be moved out. Most should be moved to Mutagen at some point.

Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/Cells/ShowSkyAnalyzerTests.cs
Elscrux: No idea why it doesn't work, probably a package update. I changed everything from ShouldEqual to ShouldEqualEnumerable now and that seems fine now.

Mutagen.Bethesda.Analyzers.Api/IAnalyzer.cs
Elscrux: I looked more into it, and I don't really want to mess with it atm because it would require a major refactoring of how engines and drivers work.

Would it be ok to merge this as a 0.0.1 and then you take a look how you want to approach this?
The main problem I see is that like I mentioned above `ByGenericTypeRecordContextualDriver ` is designed a very specific batch processing strategy of everything as opposed to just individual mod or records. And it gets more complex with `ContextualDriver` where you can't really limit anything where I suppose we'd need to run the whole thing and then catch if anything reported something for the given record x or mod y. Records themselves are also a bit problematic because in `ContextualDriver` we mostly look at the given load order and the winning record override, and not a specific override. So for records this specific driver should probably not run. For individual mods it might still be interesting though? I think it could also be interesting to expose the whole load order check functionality of  `ContextualDriver` as an api call, would also be interesting.

It gets even more convoluted with extra caches into the mix. Now we have the `ILinkCache` which are created in the `ContextualEngine` atm, but we'd actually want to support a given link cache given in the api to use instead. But when we also want extra caches this is now a situation where we have link cache / load order given from the outside to the api, and we now would need to know when to invalidate them etc, which might not be possible without extra api to support this specifically?

Another small point I want to raise is the creation of `ModPath` objects to analyze specific record frames. There we currently default to using the data directory as the base path of the mod file. I'd be good if that can also be exposed as an extra functionality to the outside because for example in my case I have an abstraction over the data directory so I can also load mods from places like MO2 mod folders, which are not the data directory. I imagine this as something that you would be able to specify from the outside via the analyzer runner builder.

As part of this update we could also make it the default case that deleted records are skipped.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Placed/Object/LocRefTypeAnalyzer.cs
Elscrux: I really should do that more right? Idk, I'm kind of paranoid from all the "Oh no this is a setter interface so I won't resolve anything from the link cache and so on" stuff, and just check use FormKeys directly a lot of the times :D

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Responses/LinksToDifferentSpeakerAnalyzer.cs
Elscrux: Yeah for sure. See my thoughts about the next steps for this here https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/pull/495#discussion_r2318497017

Elscrux:
@Noggog Thanks for the review, I think I've sorted out or addressed all you mentioned!

As a next step, although it's not fully final yet, I'd like to merge this and figure out an initial release if possible, like I mentioned here https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/pull/495#discussion_r2318497017
There will be some major engine refactoring needed to improve the concepts for this which will probably take a moment, so if we can already have this in place that'd be nice.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/UniqueNpcPlacementAnalyzer.cs
Noggog: Roger

Mutagen.Bethesda.Analyzers.Skyrim/Record/LeveledItem/ChanceNoneAnalyzer.cs
Noggog: Gotcha.  I think generally it's fine to keep things dedicated?   Im hoping these analyzers are light enough to allow for ChanceNone to be analyzed by several analyzers without needing to worry about combining them.

Combining them would be only if we thought it organized it better for our readability

Not a big deal either way

Mutagen.Bethesda.Analyzers.Skyrim/Record/Relationship/RelationshipTypeAnalyzer.cs
Noggog: PR on your PR
https://github.com/Elscrux/Mutagen.Bethesda.Analyzers/pull/2

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/PlacedObjectExtensions.cs
Noggog: I like the clarity of `if (!context.TryGetParent<IWorldspaceGetter>(out var worldspace)) return null;`, myself.   Reads easier

Mutagen.Bethesda.Analyzers.Skyrim/Record/Placed/Object/LocRefTypeAnalyzer.cs
Noggog: Haha yeah, I get that.  Just figured I'd mention.   I believe it should work if you were to use equals, but this is fine too

Mutagen.Bethesda.Analyzers.Skyrim/Record/Dialog/Topic/GiveGoldAnalyzer.cs
Noggog: Gotcha.  Yeah, not harmful.   But feels like it could go off the rails if people tried to fix too much spelling following the same patterns.   Maybe we add a more generic spellcheck system down the road when we got spare time.  Not sure

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/HeadDataExtension.cs
Noggog: Cool.  Made a card here
https://github.com/Mutagen-Modding/Mutagen/issues/640

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/PlacedObjectExtensions.cs
Elscrux: Changed it

PR #505, opened 2025-09-13
IProvideCaches mechanism

Noggog:
Gobbles up things like `VoiceTypeAssetLookupUtil` so analyzers can just request what they want and the engine can provide and also purge the caches as appropriate more easily in the future


Elscrux (approved):
I think this looks great as it is!

Big open question though is how we invalidate those caches, or if we do it at all (might also be that we create a whole new engine and drivers every time we want to work on a different state).
The api could easily expose some method to notify that some mod has changed. The cli would just be started again. An application with a UI would probably use the api functions so that's already covered too.

Noggog:
I think generally it's easier to just invalidate caches if anything changed.  Maybe we'd get a fancier setup later, but I think that's a stretch micro-optimization goal, and could be slid in whenever we feel like we have the time to execute on it

Elscrux:
> I think generally it's easier to just invalidate caches if anything changed. Maybe we'd get a fancier setup later, but I think that's a stretch micro-optimization goal, and could be slid in whenever we feel like we have the time to execute on it

You mean users invalidate caches when something is changed? Because I think we can't do that automatically from the Mutagen side. Or at least we have nothing that does this atm.
Currently you cannot do anything to invalidate the caches apart from creating a whole new analyzer pipeline.

Noggog:
Yeah, we dont have the mechanisms for it yet.   Im just saying that if/when we do, we can just substitute them in fairly cleanly now with the engine having more control over cache dissemination

PR #506, opened 2025-09-15
Analyzer runner

Noggog:
Changes the analyzer runner to not take a specific record/mod anymore, an instead just analyze whatever it's injected with from the builder


Mutagen.Bethesda.Analyzers.Cli/Mutagen.Bethesda.Analyzers.Cli.csproj
```diff
@@ -5,6 +5,7 @@
     <ItemGroup>
         <ProjectReference Include="..\Mutagen.Bethesda.Analyzers.Engine\Mutagen.Bethesda.Analyzers.Engine.csproj" />
         <ProjectReference Include="..\Mutagen.Bethesda.Analyzers.Skyrim\Mutagen.Bethesda.Analyzers.Skyrim.csproj" />
+        <ProjectReference Include="..\Mutagen.Bethesda.Analyzers\Mutagen.Bethesda.Analyzers.csproj" />
```
Elscrux: Do we want this dependency on the api project (for currently just one Autofac module)? In the future I guess we could possibly also start using the api to call the cli, though the extra run config settings in `RunAnalyzersCommand` are probably more involved, and not something the api supports atm?

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
```diff
@@ -1,90 +1,210 @@
...
+        builder
+            .RegisterInstance(new GameReleaseInjection(_gameRelease))
+            .AsImplementedInterfaces();
+
+        builder
+            .RegisterInstance(_loadOrder)
+            .AsImplementedInterfaces();
+
+        builder
+            .RegisterInstance(_linkCache)
```
Elscrux, line 275: In case we want api side cache validation we'll need to be cautious about this stuff because we'd need a different container with a different instance when this has changed.

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
```diff
@@ -1,90 +1,210 @@
...
+        if (_dataDirectory != null)
+        {
+            builder
+                .RegisterInstance(new DataDirectoryInjection(_dataDirectory.Value))
+                .AsImplementedInterfaces();
+        }
+
+        var cont = builder.Build();
+
+        return cont.Resolve<AnalyzerRunner>();
```
Elscrux: This is never registered

Mutagen.Bethesda.Analyzers/IAnalyzerRunner.cs
```diff
@@ -1,222 +1,85 @@
...
 
     internal AnalyzerRunner(
-        IFileSystem fileSystem,
-        GameRelease gameRelease,
-        ILinkCache linkCache,
-        ILoadOrderGetter<IModListingGetter<IModGetter>> loadOrder,
-        TopicConfig topicConfig,
-        Severity minimumSeverity,
-        INumWorkThreadsController numWorkThreadsController)
+        ILifetimeScope lifetimeScope)
```
Elscrux: The internal constructor won't work with Autofac

Mutagen.Bethesda.Analyzers.Engine/Drivers/IsolatedDriverParams.cs
```diff
@@ -10,20 +11,29 @@ public class IsolatedDriverParams
...
         LinkCache = linkCache;
         ReportDropbox = reportDropbox;
         TargetMod = targetMod;
-        TargetModPath = modPath;
+        FileParams = fileParams;
         CancellationToken = cancellationToken;
     }
 }
+
+public class IsolatedDriverFileParams(
```
Elscrux, line 33: Next to be able to have a custom file system, it might also be relevant to be able to pass in mod paths with different data directories. Currently you can input a custom data directory and run it all from there, but I for example also have a use case where I get mods from various MO2 folders which I all put into a game environment. This use case wouldn't be supported atm.

That would mean a mod path which is a data relative path isn't sufficient to locate a mod anymore, and you'd need to be able to target all mods individually. Probably something we don't need for the first release but it would be good to offer support for this in the future? Maybe something for an issue?

Mutagen.Bethesda.Analyzers/IAnalyzerRunner.cs
```diff
@@ -1,222 +1,85 @@
...
-
-    /// <summary>
-    /// Analyze a major record for issues.
-    /// </summary>
-    /// <param name="record">Major record to analyze</param>
-    /// <returns>Analysis results for topics found in the record</returns>
-    Task<IEnumerable<AnalyzerResult>> Analyze<TMajorRecord>(TMajorRecord record)
-        where TMajorRecord : IMajorRecordGetter;
+    /// <returns>Analysis results for topics found in the run</returns>
+    IAsyncEnumerable<AnalyzerResult> Analyze();
```
Elscrux, line 11: I think Analyze(mod) or Analyze(Record) would still be quite useful, because in a scenario where you want to analyze like these 100 records, you'd have to create 100 analyzer runners atm right?

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
```diff
@@ -1,90 +1,210 @@
...
+    }
+
+    public AnalyzerRunnerBuilder WithLinkCache(ILinkCache linkCache)
+    {
+        return new AnalyzerRunnerBuilder(
+            _gameRelease,
+            linkCache);
+    }
+
+    public AnalyzerRunnerBuilder WithLoadOrder(ILoadOrder<IModListingGetter<IModGetter>> loadOrder)
```
Elscrux: I think this might also need a variant where you're able to pass in ILoadOrder<IModListingGetter<TModGetter>> with a generic TModGetter as this is not possibel atm.

```cs
var gameEnvironment = GameEnvironment.Typical.Skyrim(SkyrimRelease.SkyrimSE);

var analyzer = AnalyzerRunnerBuilder.Create(GameRelease.SkyrimSE)
    .WithLoadOrder(gameEnvironment.LoadOrder) // error
```

Elscrux (commented):
There are some issues with DI and I had some additional comments about possible changes, but apart from that it looks pretty good!

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
Noggog: Can you elaborate on what you mean a bit?   Right now we're offering a builder where the user can give the linkcache to be used.   If the linkcache changes (because the input mods change), then a new builder given a new linkcache will be needed.  I think that's just how it is with the API we're offering with this builder?

Elscrux: Oh, yeah we'd also need to change the builder to something like an observable or so if we want to allow users to update the analyzer runner. Or we require everyone to rebuild their runner every time something changes.

Noggog: I think for now we're requiring people to redo their runner every time.    I think if we offer some crazy dynamic system in the future, we can just offer a new builder options that takes in IObservable of something and change the internals appropriately?

Mutagen.Bethesda.Analyzers.Engine/Drivers/IsolatedDriverParams.cs
Noggog: Hmm.  The only reason Data folder is needed at all is for the RecordFrame analyzers, which want to open a fresh stream to the mod's file.   They use the Data folder + ModKey to find the mod file to read in.

We could delete the Data folder concept entirely, and have them find out the mod paths a different way

Let's maybe do as a followup PR just to get this one in

Mutagen.Bethesda.Analyzers/IAnalyzerRunner.cs
Noggog: We can add these extra calls in, but I'd like to do it as a followup, as it might require other refactors

Mutagen.Bethesda.Analyzers.Engine/Drivers/IsolatedDriverParams.cs
Elscrux: A followup PR should be fine, yeah! Do we open an issue for that?

Noggog: [Right here in my pocket](https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/issues/507)

Mutagen.Bethesda.Analyzers.Engine/Autofac/DynamicAnalyzerModuleLoader.cs
```diff
@@ -0,0 +1,83 @@
...
+using Autofac;
+using Mutagen.Bethesda.Analyzers.SDK;
+using Module = Autofac.Module;
+
+namespace Mutagen.Bethesda.Analyzers.Autofac;
+
+/// <summary>
+/// Service responsible for dynamically loading game-specific analyzer modules at runtime
+/// </summary>
+internal static class DynamicAnalyzerModuleLoader
```
Elscrux, line 11: Could we expand this in the future to load external analyzer libraries?

Mutagen.Bethesda.Analyzers/IAnalyzerRunner.cs
Noggog: https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/issues/508
https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/issues/509
https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/issues/510

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
```diff
@@ -1,90 +1,322 @@
...
+            _customAnalyzers = combined
+        };
+    }
+
+    public AnalyzerRunnerBuilder WithBlacklistedMods(params ModKey[] modKeys)
+    {
+        return WithBlacklistedMods((IEnumerable<ModKey>)modKeys);
+    }
+
+    public AnalyzerRunnerBuilder WithBlacklistedMods(IEnumerable<ModKey> modKeys)
```
Elscrux, line 240: Seeing as we're working in dotnet9 we can also just use params IEnumerable<ModKey> here I think. Same for the WithAnalyzers method above.

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
```diff
@@ -1,90 +1,322 @@
...
+        {
+            builder
+                .RegisterInstance(new DataDirectoryInjection(_dataDirectory.Value))
+                .AsImplementedInterfaces();
+        }
+
+        // Register custom analyzers with distinctness by type
+        var distinctAnalyzers = _customAnalyzers
+            .GroupBy(a => a.GetType())
+            .Select(g => g.First());
```
Elscrux: Is this not basically DistinctBy() and we could use that instead?

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
```diff
@@ -1,90 +1,322 @@
...
+        // Register blacklisted mods provider
+        builder.RegisterInstance(new BuilderBlacklistedModsProvider(_blacklistedMods))
+            .As<IBlacklistedModsProvider>()
+            .SingleInstance();
+
+        var cont = builder.Build();
+
+        var factory = cont.Resolve<AnalyzerRunner.Factory>();
+
+        return factory(_workDropoff, _numWorkThreads);
```
Elscrux, line 308: Just curious, what's the reasoning behind keeping those specifically out of a the container, and passing them manually?

Elscrux (approved):
Overall this looks good to me. I haven't looked at everything very in depth but what I saw all made sense for me and it all looked like what we wanted to go for.

Mutagen.Bethesda.Analyzers.Engine/Autofac/DynamicAnalyzerModuleLoader.cs
Noggog: Eh, potentially.  The main goal of this is to -only- load the skyrim ones, so the fallout analyzers arent present at all if not needed, for ex.   Keep it clean.   Also, letting the users only deploy with the skyrim libraries if that's all they're working with on their project.

The main problem with the external libraries that this doesn't address is the mismatch potential of versions.   With this solution as it is, it's highly unlikely someone would mix `Mutagen.Bethesda.Analyzers` core library + `Mutagen.Bethesda.Analyzers.Skyrim` from a different version setup.

But it would be very likely that the external analyzer libraries were built with different mutagen versions, and that would explode.  That's the main hurdle with external analyzer libs

Noggog: Synthesis has a similar problem, which is why it runs all the patchers as CLIs.   It's near impossible to load all the patchers into memory and run them from one process

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
Noggog: Oh?  Not remembering what you're referring to offhand

Mutagen.Bethesda.Analyzers.Engine/Autofac/DynamicAnalyzerModuleLoader.cs
Elscrux: Yeah versions are a beast to tackle for sure. Interesting insight! I assume the Synthesis knowledge will come in handy here!

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
Elscrux: https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-13#params-collections

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
Noggog: Good question.   It's mainly due to an interaction of the runner being able to be run multiple times, the runner making our own workengine if not provided one, and the results not easily accessible without registering a custom handler.

All that together means we can set up most of the infrastructure once in the builder registering things into the container as normal.    But a few things need to be set up with new instances every run.

If we were to register the workengine the same way, we would need to start the consumer somewhere. And also stop it.   There's no way to know when someone is "done" using a runner and will never call Run again, and so we should stop the consumer's threads.

As such, the current system makes a new WorkEngine every time run is called, so it can start/stop it during just that run.    There's a few ways this could be done, but right now the system just passes along the bits it would need to make one to the runner so it can make one every time it gets called.

Might be a better route if i registered things differently and used transient settings better, but it does the job

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
Elscrux: So this also works now
```cs
     public AnalyzerRunnerBuilder WithBlacklistedMods(params IEnumerable<ModKey> modKeys)
```

And we don't need the extra ModKey[] method

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
Elscrux: Oh I see, so yeah that interaction of instances per run and per builder definitely makes sense.
Though for this, the work dropoff is created either per run or per builder depending if you add it via `WithDropoff` or not. I assume adding your custom work dropoff which is used globally has a benefit for some use cases?

Noggog: If it's a poweruser, they'd probably like to put in their own WorkDropoff for the whole app.    If they're telling their users they're gonna do work with only 4 threads maximum, that should be for the whole app, and they can control that better by providing their own.

Synthesis has a num threads setting for this in its UI that its wired up to a work dropoff for all the patcher building.  If it it were to ever use Analyzers, it would plop the one it already has into the builder for it to use and throw its work on the pile

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
Noggog: Ah, awesome!  I should update a good few things with that once net8 is purged in probably nov

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
Elscrux: I think I got it, that's pretty nice!

Mutagen.Bethesda.Analyzers/AnalyzerRunnerBuilder.cs
Elscrux: This project is already on dotnet9 only but yeah should be good to get that kind of stuff into Mutagen too!

PR #525, opened 2026-01-05
Cli container fixes + more analyzers

Elscrux:


Mutagen.Bethesda.Analyzers.Cli/Mutagen.Bethesda.Analyzers.Cli.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
         <OutputType>Exe</OutputType>
+        <CETCompat>false</CETCompat>
```
Noggog, line 4: Ye ol Mo2 curse 🙈

Mutagen.Bethesda.Analyzers.Engine/Drivers/Records/ByGenericTypeRecordContextualDriver.cs
```diff
@@ -36,26 +36,28 @@ public async Task Drive(ContextualDriverParams driverParams)
...
-                var param = new ContextualRecordAnalyzerParams<TMajor>(
-                    driverParams.LinkCache,
-                    driverParams.LoadOrder,
-                    listing.Mod.ModKey,
-                    rec,
-                    driverParams.ReportDropbox,
-                    driverParams.ProvideCaches);
-                return _contextualRecordAnalyzers.Select(analyzer =>
+            await Task.WhenAll(listing.Mod.EnumerateMajorRecords<TMajor>()
+                .Where(x => !x.IsDeleted)
```
Noggog, line 40: So we're saying that if a record is deleted, NO contextual analyzer can take a peek at it?   Seems like it'd be the analyzers that would decide if they want to consider deleted records and how?

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/CellExtensions.cs
```diff
@@ -34,6 +35,22 @@ public static IEnumerable<ILocationGetter> GetAllLocations(this ICellGetter cell
         return cellLocations;
     }
 
+    /// <summary>
+    /// Estimates if a cell is just a testing cell that can be ignored.
+    /// A testing cell is always an interior cell, and has no special setup.
+    /// </summary>
+    /// <param name="cell">Cell to check</param>
+    /// <returns>True if the cell is likely a testing cell</returns>
+    public static bool IsTestingCell(this ICellGetter cell)
```
Noggog, line 44: Probably fine, but I think i'd prefer building up a good default config that ignores known testing cells?  Rather than logic that tries to estimate.  Maybe the estimation logic can be used to find the testing cells to put on the ignore list, but not actually be used?   Just thinkin out loud

Mutagen.Bethesda.Analyzers.Skyrim/SkyrimAnalyzerModule.cs
```diff
@@ -20,5 +21,11 @@ protected override void Load(ContainerBuilder builder)
             .AsSelf()
             .AsImplementedInterfaces()
             .SingleInstance();
+
+        builder.RegisterAssemblyTypes(typeof(VoiceTypeAssetLookupProvider).Assembly)
```
Noggog, line 25: This pattern is fine enough, where we register each folder as we see fit.

Lately, though, i've been flipping it around where I do something like:
```
Services/
   Singletons/
     /AnyOrganizationWeLike
   Transients/
     /AnyOrganizationWeLike
OtherThings/
OtherOtherThings/
```
Where if you want to get registered in the module automatically, you put yourself under either singleton/transient folder appropriately.   

Keeps the module simple, and feels nice and organized to me, so far.

Anyway, something to consider.  Not a big deal either way

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/NoCleanupScriptAnalyzer.cs
```diff
@@ -36,20 +50,62 @@ public void AnalyzeRecord(ContextualRecordAnalyzerParams<INpcGetter> param)
         // Skip NPCs using templates for scripts
         if (npc.Configuration.TemplateFlags.HasFlag(NpcConfiguration.TemplateFlag.Script)) return;
 
+        // Essential NPCs can't die and don't need cleanup scripts
+        if (npc.Configuration.Flags.HasFlag(NpcConfiguration.Flag.Essential)) return;
+
+        // Children can't die and don't need cleanup scripts
+        var race = npc.Race.TryResolve(param.LinkCache);
+        if (race is not null && race.Flags.HasFlag(Race.Flag.Child)) return;
```
Noggog, line 58: Don't they have a killable children mod? haha

Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
```diff
@@ -65,8 +69,40 @@ private static IContainer GetContainer(RunAnalyzersCommand command)
...
+            .RegisterInstance(gameEnvironment.LinkCache)
+            .AsImplementedInterfaces();
+
+        var workDropoff = new WorkDropoff();
+        builder
+            .RegisterInstance(workDropoff)
+            .AsImplementedInterfaces();
+
+        builder
+            .RegisterInstance(new WorkConsumer(0, workDropoff, workDropoff))
```
Noggog: I think this needs to be started?  Are you sure we dont want 1 thread at least?  Perhaps you want InlineWorkDropoff?  But i imagine we'll want a threaded dropoff, but just with more control over how many to spawn for the run

Mutagen.Bethesda.Analyzers.Skyrim/Record/Npc/Unique/NoCleanupScriptAnalyzer.cs
Elscrux: There is a mod for it but I think we want to create this analyzer for the base game, and not have sort of false positive or redundant errors for children where it's not a problem unless you install a mod like that.

Mutagen.Bethesda.Analyzers.Cli/Mutagen.Bethesda.Analyzers.Cli.csproj
Elscrux: Yeah :(

Mutagen.Bethesda.Analyzers.Engine/Drivers/Records/ByGenericTypeRecordContextualDriver.cs
Elscrux: Yeah in the future we should build this in a way where the analyzers could decide. So far I didn't see any analyzer that needed this so I built this like that for now.

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/CellExtensions.cs
Elscrux: I think it's tricky because you want to make this possible to be used for any mod, so I wouldn't restrict it just to Skyrim. However, this is still useful to filter out some of the records you wouldn't want to check.

On another note though, we already have ideas to add a way to specify blacklisted record though because that will be useful to filter out any false positives people want to ignore, if they choose to. Still need to do that though.

Mutagen.Bethesda.Analyzers.Skyrim/SkyrimAnalyzerModule.cs
Elscrux: Interesting, yeah we could try something like that. My current practice for my own projects is still to register everything manually just because I can better keep track of stuff, but as far as registering by assembly, this seems like a good way to do it.

Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
Elscrux: I just added something to make it work somehow, I am not too familiar with the concepts here.

Though I can't just change `WorkDropoff` to `InlineWorkDropoff` it seems because `InlineWorkDropoff` doesn't implemented `IWorkQueue`. Do you have an idea how to best resolve this?

Elscrux: Ok, no it seems like the WorkConsumer is needed and to be created it needs a `IWorkQueue` which only implemented by `WorkDropoff` and not `InlineWorkDropoff`, so this won't work.

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/CellExtensions.cs
Noggog: My hesitance to this as it is is that its cutting off future configuration.

Without the IsTestingCell test, you're getting lots of false positives.   Sure.   But you can blacklist those records from firing the errors, and now it's clean.  But you can't do the reverse.  How would someone say "i dont want it to hide things if it thinks its a test cell -> i want to see the results anyway"?   They can't

You can add blacklists via the config to silence the false positives... but you can't do the reverse and rip out the `if (cell.IsTestingCell()) return;` short circuit via the config.  So the user can't get it to be extra careful when they want it.

`if (cell.IsExteriorCell()) return;` is fine for NorthMarkerAnalyzer, for ex, cuz that's more of a certainty:  It's known to not matter if exterior (i imagine), and the test for it is known.

I think i'd rather lean on filling out the testing cell blacklist?  Or.. looking at other alternatives.  Could add a TestCell keyword tag, and if it's got that keyword -> skip it.  Then some patcher can be the one to determine which are test cells and stamp them appropriately, or the user can stamp it with the tag manually as they decide.  

We'd maybe add a feature for blacklists to blacklist a keyword, then, and we'd be in the best of both worlds situation i think.

But having a hardcoded but best guess fuzzy function that you can't config away feels wrong.

Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
Noggog: IWorkDropoff is the important interface.   The things that want work to be done import that, and drop off the work to be done, and optionally await the signal that it's done.

InlineWorkDropoff is the "actually just do it immediately" implementation.  When people drop off the work to be done, it just does it right then and there.  That's it.  No other interfaces/implementations involved.

----------
But, you can opt to use more complex implementations.   

The other typical one involves two concrete objects:
- A WorkConsumer which knows how many threads to make to do work.  It looks at IWorkQueue to pull things to do off and then executes them.   This object needs to be .Start()ed before it'll start pulling off the queue.
- A queue object that implements IWorkDropoff for people to put work on the queue, and IWorkQueue for the WorkConsumer to grab work off

-----
I would register WorkConsumer and WorkDropoff as implemented interfaces.  During startup, resolve out IWorkConsumer and run its start call.   You'll also need to register a INumWorkThreadsController to determine how many threads the consumer should use.   You an register the `NumWorkThreadsConstant` if you just want it set to a specific number, `NumWorkThreadsUnopinionated` to let it choose CPU count automatically, or your own INumWorkThreadsController if you want it hooked to a UI to control how many threads to use, for ex.

analyzers will be pulling in IWorkDropoff and dropping things onto the queue, which gets pulled off by the consumer and completed on some background thread

Elscrux: Ok I registered `INumWorkThreadsController` now too, and I removed the WorkConsumer setup. I didn't realize that the WorkConsumer part was something that not the analyzer framework needs inherently, but something that was custom setup for the Cli as well, so I just removed that work and used to the InlineWorkDropoff like you said.

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/CellExtensions.cs
Elscrux: I fully agree we need a blacklist system, but this should be used for other parts. I think the analyzers should be able to limit their scope to a certain part of records too. There is no situation where testing cells need a north marker so they should also not be raised for that.

I see what you're getting at, but from my view it's a filter to help you so you don't need to go through an annoyingly long list of problem which don't matter so for the obvious cases we can already filter it out. Just like we would filter it out for settlement cell analyzers that should have a lock list. You could also run this on all interiors, but it will only matter for the settlement cells and you have a lot of false positives. And in this case it's relevant for all cells that are actually used, and not testing cells, so we filter those out at the start.

Noggog: When it's completely known and deterministic, I'd totally agree an analyzer should be able to short circuit.   If an analyzer knows it only applies to external cells, for ex, it can literally just check a flag, know it's external for certain, and skip.

But this here is instead a guess.  There is no actual data that represents "I am test cell".  So we just estimate based on clues.   And it's a guess the user can't say "hey, you were wrong in your estimation, stop considering this cell a test cell and analyze it for real, please!".  

That's the main distinction, to me.    When it's a guess, we need an escape hatch for when that guess is wrong.  The escape hatch normally is the config.  When the user thinks it's wrong on something, they can silence an error for a particular record.   Maybe we can think on it more and change later.

Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
Noggog: Should work, it'll just be single threaded until we add some more upgrades

Noggog (approved):

PR #526, opened 2026-01-21
Adjust registration for work engine on CLI runs

Noggog:
NumThreads parameter was already registered.  Not sure why I wouldn't have registered the work engine as well back then.

Give this a spin and let me know if it works as expected!


Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
```diff
@@ -22,15 +22,19 @@ namespace Mutagen.Bethesda.Analyzers.Cli;
...
 public static class RunAnalyzers
 {
+    private record Bootstrap(ContextualAnalyzerEngine Engine, IWorkConsumer WorkConsumer);
+
     public static async Task<int> Run(RunAnalyzersCommand command)
     {
         var container = GetContainer(command);
 
-        var engine = container.Resolve<ContextualAnalyzerEngine>();
+        var bootstrap = container.Resolve<Bootstrap>();
```
Elscrux, line 33: `Bootstrap` is not registered

Elscrux (changes requested):

Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
Noggog: Roger.  Will write some more tests to have caught this

PR #529, opened 2026-01-25
fix: handle data folder when --DataFolder is provided

### LINKED ISSUE #527: Some analyzers don't run
I am keeping an eye on this project and I noticed there were some updates recently, so I cloned the latest `dev` (d99a5dcbcd47c81bafa0344dfc3f4bd32f9d9b73 at the time of writing) to try it out locally. I tried running that version with some of my plugins and I noticed that some reports that were previously present are now missing when running the latest version. I noticed this issue with `ObjectWindowFilterAnalyzer`, but I suspect there may be more.

The record was reported fine in 343c554bd9e780fbf0a26e045445b972d26e670a. Note: that commit in particular didn't produce a working build as-is: it returned an exception at runtime due to some AutoFac dependency issues, which I fixed locally (I can provide a git diff that will fix that commit if necessary for reproducing the issue), but apart from that it was working fine.

## Reproduction steps

I have attached to this issue a simple .esp for Skyrim SE **(here: 
[mutagen_analyzers_quest_test.zip](https://github.com/user-attachments/files/24836851/mutagen_analyzers_quest_test.zip)**) which only contains a Quest record (`MutagenTestQuest`) without an object window filter. The plugin has `Skyrim.esm` as its only dependency.

The Quest record should trigger https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/blob/dev/Mutagen.Bethesda.Analyzers.Skyrim/Record/Quest/ObjectWindowFilterAnalyzer.cs#L8, but nothing is reported by the CLI.

Image

I ran the analyzers like this, black-listing `skyrim.esm` to avoid noise:

```
Mutagen.Bethesda.Analyzers.Cli.exe run-analyzers --GameRelease SkyrimSE --BlacklistedMods skyrim.esm --DataFolder ".\Data" --LoadOrder skyrim.esm,mutagen_analyzers_quest_test.esp
```

**Output before:**

```
[A253][SUG] No Object Window Filter
   mutagen_analyzers_quest_test.esp -> 000D62:mutagen_analyzers_quest_test.esp MutagenTestQuest
   Quest has no Object Window Filter
```

**Output with latest `dev`:**

Nothing is printed by the CLI. The program exits without printing anything.

**Expected output:**

 same as 343c554bd9e780fbf0a26e045445b972d26e670a (`[A253][SUG] No Object Window Filter` should be reported).

## Other findings

I tried setting a breakpoint in the `AnalyzeRecord` function and it looks like it'is never being hit.

xnyo:
`--DataFolder` argument is not passed to the game environment builder. If the argument is provided, the game's installation folder is used instead of the provided one. Fixes #527.

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
+++ b/Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
@@ -78,9 +78,15 @@ public static class RunAnalyzers
                 .Select(x => x.Trim())
                 .Select(x => ModKey.FromFileName(x));
 
-            gameEnvironment = GameEnvironmentBuilder.Create(GameRelease.SkyrimSE)
-                .WithLoadOrder(loadOrder.ToArray())
-                .Build();
+            var envBuilder = GameEnvironmentBuilder.Create(command.GameRelease)
+                .WithLoadOrder(loadOrder.ToArray());
+
+            if (command.DataFolder is not null)
+            {
+                envBuilder = envBuilder.WithTargetDataFolder(new DirectoryPath(command.DataFolder));
+            }
+
+            gameEnvironment = envBuilder.Build();
         }
 
         builder


Elscrux:
Thanks! Just one thing, I think @Noggog also included basically the same code in #526

xnyo:
> Thanks! Just one thing, I think @Noggog also included basically the same code in #526

I have just noticed that! 😅 Feel free to close the PR if you want

Noggog (approved):

Noggog:
Thanks for the PR!  Merged!

PR #546, opened 2026-03-07
Add some more analyzers

Elscrux:


Mutagen.Bethesda.Analyzers.Skyrim/Record/Scene/SceneActionOverlapAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+
+    public IEnumerable<TopicDefinition> Topics { get; } = [OverlappingActionsOfSameType];
+
+    public void AnalyzeRecord(IsolatedRecordAnalyzerParams<ISceneGetter> param)
+    {
+        var scene = param.Record;
+
+        // Group actions by actor ID
+        var actionsByActor = scene.Actions
+            .Where(a => a.ActorID is not null && a.ActorID >= 0)
```
Noggog: Probably can just slap a `.WhereNotNull()` in there rather than adding the forgiveness operator

Mutagen.Bethesda.Analyzers.GitHub/Mutagen.Bethesda.Analyzers.GitHub.csproj
```diff
@@ -5,6 +5,7 @@
     <ItemGroup>
         <ProjectReference Include="..\Mutagen.Bethesda.Analyzers.Common\Mutagen.Bethesda.Analyzers.Common.csproj" />
         <ProjectReference Include="..\Mutagen.Bethesda.Analyzers.Engine\Mutagen.Bethesda.Analyzers.Engine.csproj" />
+        <ProjectReference Include="..\Mutagen.Bethesda.Analyzers.Skyrim\Mutagen.Bethesda.Analyzers.Skyrim.csproj" />
```
Noggog, line 8: This add needed?  If so, maybe add all the games?

Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
```diff
@@ -24,52 +24,139 @@ public class ConditionAnalyzer : IContextualRecordAnalyzer<ISkyrimMajorRecordGet
...
+                        {
+                            param.AddTopic(GetCurrentTimeConditionWithAndOnDayBreak.Format());
+                        }
+                    }
+
+                    break;
+                }
+                case IGetCrimeGoldConditionDataGetter getCrimeGold:
+                    if (condition.Data.RunsOnPlayer() && getCrimeGold.Faction.UsesLink() && getCrimeGold.Faction.Link.IsNull) {
+                        Console.WriteLine("");
```
Noggog: Console.Write seems odd?  Just a placeholder for a new topic here at some point?

Noggog (changes requested):
Mainly just the console writes that look odd

Mutagen.Bethesda.Analyzers.GitHub/Mutagen.Bethesda.Analyzers.GitHub.csproj
Elscrux: Iirc, this is needed to make sure it can find the types in the assembly. And Skyrim is the only game rn.

Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
Elscrux: Oh, thanks for noticing this! I used this to debug the topics and get a breakpoint but forgot to actually add the topic at some point. Should be fixed now.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Scene/SceneActionOverlapAnalyzer.cs
Elscrux: Problem is that the group by needs the actions, and the null check is on the actor id in the action. So I can't use Select first and then WhenNotNull although that would be super neat. Unless you have any other idea?

Noggog (approved):

PR #548, opened 2026-03-18
Created Analyzer for Discussion #547

FaffiSalus:
Created Analyzer for MagicEffect Description

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/MagicEffects/DescriptionAnalyzerTests.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/MagicEffects/DescriptionAnalyzerTests.cs
@@ -10,24 +10,27 @@ namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.IsolatedRecords.MagicEffects;
 public class DescriptionAnalyzerTests
 {
     [Theory, MutagenModAutoData]
-    public void DetectsSinglePercent(
+    public void DetectsIncorrectPercentageUsage(
         IsolatedRecordTestFixture<DescriptionAnalyzer, MagicEffect, IMagicEffectGetter> fixture)
     {
         fixture.Run(
             prepForError: rec => rec.Description = new TranslatedString(Language.English,
                 new KeyValuePair<Language, string>(Language.English, "Deals 50% damage"),
-                new KeyValuePair<Language, string>(Language.French, "Inflige 50% de dégâts"),
+                new KeyValuePair<Language, string>(Language.French, "Inflige 50%%%%%% de dégâts"),
                 new KeyValuePair<Language, string>(Language.Danish, "50%"),
-                new KeyValuePair<Language, string>(Language.German, "%50"),
-                new KeyValuePair<Language, string>(Language.Arabic, "%")),
+                new KeyValuePair<Language, string>(Language.German, "%50%"),
+                new KeyValuePair<Language, string>(Language.Arabic, "%"),
+                new KeyValuePair<Language, string>(Language.Chinese, "5%%%")),
             prepForFix: rec => rec.Description = new TranslatedString(Language.English,
                 new KeyValuePair<Language, string>(Language.English, "Deals 50%% damage"),
                 new KeyValuePair<Language, string>(Language.French, "Inflige 50%% de dégâts"),
                 new KeyValuePair<Language, string>(Language.Danish, "50%%"),
-                new KeyValuePair<Language, string>(Language.German, "%%50"),
-                new KeyValuePair<Language, string>(Language.Arabic, "%%")),
+                new KeyValuePair<Language, string>(Language.German, "%%50%%"),
+                new KeyValuePair<Language, string>(Language.Arabic, "%%"),
+                new KeyValuePair<Language, string>(Language.Chinese, "5%%")),
             new[]
             {
+                DescriptionAnalyzer.MagicEffectDescriptionList,
                 DescriptionAnalyzer.MagicEffectDescriptionList,
                 DescriptionAnalyzer.MagicEffectDescriptionList,
                 DescriptionAnalyzer.MagicEffectDescriptionList,
--- a/Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/MagicEffects/DescriptionAnalyzerTests.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/MagicEffects/DescriptionAnalyzerTests.cs
@@ -1,0 +1,39 @@
+using Mutagen.Bethesda.Analyzers.Skyrim.Record.MagicEffect;
+using Mutagen.Bethesda.Analyzers.Testing.Frameworks;
+using Mutagen.Bethesda.Skyrim;
+using Mutagen.Bethesda.Strings;
+using Mutagen.Bethesda.Testing.AutoData;
+using Xunit;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.IsolatedRecords.MagicEffects;
+
+public class DescriptionAnalyzerTests
+{
+    [Theory, MutagenModAutoData]
+    public void DetectsSinglePercent(
+        IsolatedRecordTestFixture<DescriptionAnalyzer, MagicEffect, IMagicEffectGetter> fixture)
+    {
+        fixture.Run(
+            prepForError: rec => rec.Description = new TranslatedString(Language.English,
+                new KeyValuePair<Language, string>(Language.English, "Deals 50% damage"),
+                new KeyValuePair<Language, string>(Language.French, "Inflige 50% de dégâts"),
+                new KeyValuePair<Language, string>(Language.Danish, "50%"),
+                new KeyValuePair<Language, string>(Language.German, "%50"),
+                new KeyValuePair<Language, string>(Language.Arabic, "%")),
+            prepForFix: rec => rec.Description = new TranslatedString(Language.English,
+                new KeyValuePair<Language, string>(Language.English, "Deals 50%% damage"),
+                new KeyValuePair<Language, string>(Language.French, "Inflige 50%% de dégâts"),
+                new KeyValuePair<Language, string>(Language.Danish, "50%%"),
+                new KeyValuePair<Language, string>(Language.German, "%%50"),
+                new KeyValuePair<Language, string>(Language.Arabic, "%%")),
+            new[]
+            {
+                DescriptionAnalyzer.MagicEffectDescriptionList,
+                DescriptionAnalyzer.MagicEffectDescriptionList,
+                DescriptionAnalyzer.MagicEffectDescriptionList,
+                DescriptionAnalyzer.MagicEffectDescriptionList,
+                DescriptionAnalyzer.MagicEffectDescriptionList
+            }
+            );
+    }
+}
--- a/Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
@@ -1,0 +1,51 @@
+using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Skyrim;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.MagicEffect;
+
+public class DescriptionAnalyzer : IIsolatedRecordAnalyzer<IMagicEffectGetter>
+{
+    //TBD
+     public static readonly TopicDefinition MagicEffectDescriptionList = MutagenTopicBuilder.FromDiscussion(
+             547,
+             "MagicEffectDescription",
+             Severity.Suggestion)
+         .WithoutFormatting("% in Magic Effect description should be followed by a second %");
+
+    public IEnumerable<TopicDefinition> Topics { get; } = [MagicEffectDescriptionList];
+
+    public void AnalyzeRecord(IsolatedRecordAnalyzerParams<IMagicEffectGetter> param)
+    {
+        var mgef = param.Record;
+        if (mgef.Description != null)
+        {
+            string? desc = mgef.Description.String;
+            int i = 0;
+            if (desc != null)
+            {
+                while (i < desc.Length)
+                {
+                    if (desc[i] == '%')
+                    {
+                        if (desc[i + 1] != '%')
+                        {
+                            param.AddTopic(MagicEffectDescriptionList.Format());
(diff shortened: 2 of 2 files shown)


Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+             "MagicEffectDescription",
+             Severity.Suggestion)
+         .WithoutFormatting("% in Magic Effect description should be followed by a second %");
+
+    public IEnumerable<TopicDefinition> Topics { get; } = [MagicEffectDescriptionList];
+
+    public void AnalyzeRecord(IsolatedRecordAnalyzerParams<IMagicEffectGetter> param)
+    {
+        var mgef = param.Record;
+        if (mgef.Description != null)
```
Elscrux: I'd go with an early return here to reduce the amount of nesting we have in this method.

Like this:
```cs
if (mgef.Description is null) return;
```

Also a note that I'm using `is null` instead of `== null`. That's a convention in C# when you want to be sure to use the null checking because it might for some reason be that the == operator was overloaded to give `== null` a different meaning. Doesn't change anything here, just a thing for convention which might be good to adopt :)

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            string? desc = mgef.Description.String;
+            int i = 0;
+            if (desc != null)
```
Elscrux: Same thing with the early return here. Something you could even combine both of these.

```cs
if (mgef.Description?.String is null) return;
```

But what would actually be best for this because it's a translated string is that you iterate through all the different descriptions in the different languages and check if any language has this issue. For this it's also important to include the language in the topic that you report to make sure people know which language they need to fix this in.

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
+using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Skyrim;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.MagicEffect;
+
+public class DescriptionAnalyzer : IIsolatedRecordAnalyzer<IMagicEffectGetter>
+{
+    //TBD
```
Elscrux: I guess this can go

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            string? desc = mgef.Description.String;
+            int i = 0;
+            if (desc != null)
+            {
+                while (i < desc.Length)
+                {
+                    if (desc[i] == '%')
+                    {
+                        if (desc[i + 1] != '%')
```
Elscrux: What happens if the % is at the end of the string? This +1 access would be out of bounds and this would throw. We need to be sure to catch that. So either if the next char is not % or if the string has ended we want to raise this topic.

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+using Mutagen.Bethesda.Skyrim;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.MagicEffect;
+
+public class DescriptionAnalyzer : IIsolatedRecordAnalyzer<IMagicEffectGetter>
+{
+    //TBD
+     public static readonly TopicDefinition MagicEffectDescriptionList = MutagenTopicBuilder.FromDiscussion(
+             547,
+             "MagicEffectDescription",
```
Elscrux: This should be a short description, basically the same as the title that you gave the discussion. That should usually match and this is good `Single '%' in Magic Effect Description`

Elscrux (changes requested):
Thanks a lot! I had some notes :)

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
Noggog: > == operator was overloaded to give == null a different meaning

Woof, that'd be a weird one.   Not really one i'd worry about too much, myself?  If they overload it so that null so much so it behaves differently and breaks the nullability compiler... i'd say that's a naughty library?  

i think generally equality should aim to be symmetrical?  Making the overload operator so that `mgef.Description != null` wasnt symmetrical with `null != mgef.Description` is sus

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+            int i = 0;
+            if (desc != null)
+            {
+                while (i < desc.Length)
+                {
+                    if (desc[i] == '%')
+                    {
+                        if (desc[i + 1] != '%')
+                        {
+                            param.AddTopic(MagicEffectDescriptionList.Format());
```
Noggog: Maybe break here if we add the topic?  No need to keep looking for more topics to raise

If the topic had info on the positioning of the offense, then maybe it would make sense to log each one

Noggog (changes requested):
Thanks for the PR!

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
Elscrux: Yeah it should usually not be the case but I know that at least the Unity engine does this and because of that it's become a bit of a standard at least in some circles to use `is null` just to be sure.

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
FaffiSalus: Looping through languages is easy enough but I am struggling to come up with an elegant solution to add the language to the topic. Do you have any suggestions or is there an already existing Analyzer that does this I could look at?

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
FaffiSalus: It wasn't running out of bounds as the loop is for i < desc.Length but that meant i was never checking the last character of the string. I've fixed this now and added a comment because the solution is slightly confusion to look at

FaffiSalus: Nvm I was stupid, actually fixed the issue now

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,46 @@
...
+    public IEnumerable<TopicDefinition> Topics { get; } = [MagicEffectDescriptionList];
+
+    public void AnalyzeRecord(IsolatedRecordAnalyzerParams<IMagicEffectGetter> param)
+    {
+        var mgef = param.Record;
+        if (mgef.Description is null) return;
+        foreach (Language languages in Enum.GetValues(typeof(Language)))
+        {
+            string? desc = mgef.Description.Lookup(languages);
+            if ((desc is null)) continue;
```
Elscrux: There are redundant parentheses here

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,46 @@
...
+             "Single '%' in Magic Effect Description",
+             Severity.Suggestion)
+         .WithoutFormatting("% in Magic Effect description should be followed by a second %");
+
+    public IEnumerable<TopicDefinition> Topics { get; } = [MagicEffectDescriptionList];
+
+    public void AnalyzeRecord(IsolatedRecordAnalyzerParams<IMagicEffectGetter> param)
+    {
+        var mgef = param.Record;
+        if (mgef.Description is null) return;
```
Elscrux, line 21: Just a style thing that I'm mentioning because I know you're new to the language. It's not really relevant for the function of the program of course, just something that might be good to have a standardized style for. I know we have a editorconfig file in this project which should help a little for that: [https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/blob/dev/.editorconfig](https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/blob/dev/.editorconfig). Idk if you're currently using that or what IDE you're using and if that would use this editorconfig which basically codifies some best practices. Basically when your IDE supports it you press the hotkey for formatting everything and then it takes care of doing that based on the editorconfig and we should be good!

After a control flow statement like return, or continue with code after that, I'd always make an empty line to make it easy to see that there is a branch in the control flow here.

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,46 @@
...
+        foreach (Language languages in Enum.GetValues(typeof(Language)))
+        {
+            string? desc = mgef.Description.Lookup(languages);
+            if ((desc is null)) continue;
+            int i = 0;
+            while (i < desc.Length)
+            {
+                if (desc[i] == '%')
+                {
+                    if (i+1 >= desc.Length || desc[i + 1] != '%')
```
Elscrux: Also style thing, I'd prefer to use spaces across operators like +, you did it here once and once you didn't.

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.MagicEffect;
+
+public class DescriptionAnalyzer : IIsolatedRecordAnalyzer<IMagicEffectGetter>
+{
+     public static readonly TopicDefinition MagicEffectDescriptionList = MutagenTopicBuilder.FromDiscussion(
+             547,
+             "Single '%' in Magic Effect Description",
+             Severity.Suggestion)
+         .WithoutFormatting("% in Magic Effect description should be followed by a second %");
```
Elscrux: This still doesn't have the language in the topic. This should be mentioned in there to make sure people know which language they need to fix.

Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/MagicEffects/DescriptionAnalyzerTests.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.IsolatedRecords.MagicEffects;
+
+public class DescriptionAnalyzerTests
+{
+    [Theory, MutagenModAutoData]
+    public void DetectsSinglePercent(
+        IsolatedRecordTestFixture<DescriptionAnalyzer, MagicEffect, IMagicEffectGetter> fixture)
+    {
+        fixture.Run(
+            prepForError: rec => rec.Description = new TranslatedString(Language.English,
```
Elscrux, line 17: For this test it might be good to test some other cases where it shouldn't report because it has two %%, or more. Some edge cases, like the one we fixed now where the single % is at the end.

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
```diff
@@ -0,0 +1,46 @@
...
+
+    public IEnumerable<TopicDefinition> Topics { get; } = [MagicEffectDescriptionList];
+
+    public void AnalyzeRecord(IsolatedRecordAnalyzerParams<IMagicEffectGetter> param)
+    {
+        var mgef = param.Record;
+        if (mgef.Description is null) return;
+        foreach (Language languages in Enum.GetValues(typeof(Language)))
```
Elscrux: Btw, when you look at something like this, it also shows you an easy way to iterate through a translated string where you don't even have to do a lookup call manually.

https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/blob/dev/Mutagen.Bethesda.Analyzers.Skyrim/Record/Cell/InconsistentCellNameAnalyzer.cs

Elscrux (changes requested):
Nice work, I got some more little notes, mostly styling :)

Mutagen.Bethesda.Analyzers.Skyrim/Record/MagicEffect/DescriptionAnalyzer.cs
Elscrux: https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/blob/dev/Mutagen.Bethesda.Analyzers.Skyrim/Record/Cell/InconsistentCellNameAnalyzer.cs

this should help :)

Elscrux (approved):
Thank you, I think I don't have any more comments! :)

Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/MagicEffects/DescriptionAnalyzerTests.cs
```diff
@@ -0,0 +1,42 @@
...
+using Mutagen.Bethesda.Strings;
+using Mutagen.Bethesda.Testing.AutoData;
+using Xunit;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.IsolatedRecords.MagicEffects;
+
+public class DescriptionAnalyzerTests
+{
+    [Theory, MutagenModAutoData]
+    public void DetectsIncorrectPercentageUsage(
```
Noggog, line 13: Love the unit tests!

Noggog (approved):

PR #550, opened 2026-03-22
Created new Analyzer for Discussion #549

FaffiSalus:
New Analyzer for Factions with Blacklist VendorBuySellList that don't have the NotBuySell flag set

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim/Record/Faction/FactionNotBuySellAnalyzer.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/Record/Faction/FactionNotBuySellAnalyzer.cs
@@ -2,12 +2,19 @@ using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
 using Mutagen.Bethesda.Analyzers.SDK.Topics;
 using Mutagen.Bethesda.Plugins;
 using Mutagen.Bethesda.Skyrim;
-using Noggog;
 
 namespace Mutagen.Bethesda.Analyzers.Skyrim.Record.Faction;
 
 public class FactionNotBuySellAnalyzer : IIsolatedRecordAnalyzer<IFactionGetter>
 {
+    //build FormKey HashSet for VendorBuyLists that should have NotSellBuy set to true
+    private static readonly HashSet<FormKey> VendorBuySellLists =
+    [
+        FormKeys.SkyrimSE.Skyrim.FormList.VendorItemsMisc.FormKey,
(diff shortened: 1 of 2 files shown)


Mutagen.Bethesda.Analyzers.Skyrim/Record/Faction/FactionNotBuySellAnalyzer.cs
```diff
@@ -0,0 +1,57 @@
...
+        var faction = param.Record;
+
+        if (!faction.IsVendor()) return;
+
+        if (faction.VendorValues is null) return;
+
+        if (faction.VendorBuySellList.FormKey == FormKey.Null) return;
+
+        //build FormKey Array for VendorBuyLists that should have NotSellBuy set to true
+        FormKey[] vendorBuySellLists =
```
Noggog: Since this doesn't change, we can put it as a static field so it only gets made once instead of every time the analyzer runs

Mutagen.Bethesda.Analyzers.Skyrim/Record/Faction/FactionNotBuySellAnalyzer.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        bool relevantFaction = false;
+        foreach (FormKey formKey in vendorBuySellLists)
+        {
+            if (faction.VendorBuySellList.FormKey == formKey)
+            {
+                relevantFaction = true;
+            }
+        }
+
```
Noggog: If we upgrade vendorBuySellLists to a HashSet<FormKey>, we can just do a contains call
`if (!vendorBuySellLists.Contains(faction.VendorBuySellList)) return;`

Mutagen.Bethesda.Analyzers.Skyrim/Record/Faction/FactionNotBuySellAnalyzer.cs
```diff
@@ -0,0 +1,57 @@
...
+        if (!relevantFaction) return;
+
+        if (faction.VendorValues.NotSellBuy == false)
```
Noggog: I'd prefer `if (!faction.VendorValues.NotSellBuy)`

Noggog (changes requested):

Mutagen.Bethesda.Analyzers.Skyrim/Record/Faction/FactionNotBuySellAnalyzer.cs
FaffiSalus: I've never used one, i'll look into it!

Noggog (approved):

PR #552, opened 2026-03-24
Allow GetStage == 0 even if quest does not have a stage 0

kieranknowles1:
GetStage returns 0 for quests that have not had any stages set, such as if they are yet to start.


Elscrux (approved):
Thanks! Yeah this is a valid point. I'll also adapt the discussion for this to reflect this change

PR #555, opened 2026-03-27
Create analyzer for discussion #554

kieranknowles1:
Analyzer for Non-optional Alias Forced to None

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/Quests/AliasForcedNoneAnalyzerTest.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/Quests/AliasForcedNoneAnalyzerTest.cs
@@ -1,0 +1,189 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Mutagen.Bethesda.Analyzers.Skyrim.Record.Quest;
+using Mutagen.Bethesda.Analyzers.Testing.Frameworks;
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Skyrim;
+using Mutagen.Bethesda.Testing.AutoData;
+using Xunit;
+
+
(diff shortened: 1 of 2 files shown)


Noggog (approved):
Heya!  Thanks for the PRs!

I think i avoided requiring it early on for @Elscrux as it was so early, but I think it would be good policy to start generally requiring some basic unit tests

We have a decent suite for making this not too painful, which you can see here:
https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/blob/dev/Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/Cells/CellUnderscoreAnalyzerTests.cs

Basically it just requires:
1) What analyzer class is in play?
2) What record type is in pay?
3) What is the topic(s) is expected to be raised in the test?
4) What should the record look like so it triggers the topic?
5) What minimal changes would I do on it so the raised topic goes away?

Should hopefully allow for unit testing that isn't too much of a PITA to write?

Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/Quests/AliasForcedNoneAnalyzerTest.cs
```diff
@@ -0,0 +1,189 @@
...
+using System.Threading.Tasks;
+using Mutagen.Bethesda.Analyzers.Skyrim.Record.Quest;
+using Mutagen.Bethesda.Analyzers.Testing.Frameworks;
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Skyrim;
+using Mutagen.Bethesda.Testing.AutoData;
+using Xunit;
+
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.IsolatedRecords.Quests
```
Noggog, line 14: Not a big deal, but in recent C# versions we can do this

`namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.IsolatedRecords.Quests;`

and not nest the rest of the file

Noggog (approved):

PR #556, opened 2026-03-28
Adding UnitTests

FaffiSalus:
Added more missing UnitTests


Mutagen.Bethesda.Analyzers.Skyrim.Tests/IsolatedRecords/Cells/Exterior/EncounterZoneAnalyzerTest.cs
```diff
@@ -0,0 +1,31 @@
+using Mutagen.Bethesda.Analyzers.Skyrim.Record.Cell.Exterior;
+using Mutagen.Bethesda.Analyzers.Testing.Frameworks;
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Skyrim;
+using Mutagen.Bethesda.Testing.AutoData;
+using Xunit;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.IsolatedRecords.Cells.Exterior;
+
+public class EncounterZoneAnalyzerTest
```
Noggog, line 10: Might be good to set explicitly that the cell is exterior in the test

`IsExteriorCell` checks that IsInteriorCell is false
While that situation might be true by default, it's good to set it in the test so we know it's an important aspect of whether it passes/fails.  That, and default values changing for whatever reason would mess things up

Noggog (changes requested):

Noggog (approved):

PR #557, opened 2026-03-28
Fix checking for errors after prepareForFix

kieranknowles1:
dropOff was replaced with a new object without updating the destination of reports. This caused prepareForFix to always succeeed, even if there were still errors.


kieranknowles1:
Tests are currently failing as AliasIDAnalyzerTest creates new aliases during prepForFix rather than updating the existing ones

Noggog (approved):

PR #562, opened 2026-04-11
Fix false positive in UniqueNpcPlacementAnalyzer

kieranknowles1:
Analyser was checking for PlacedNpcs referencing a unique NPC rather than PlacedNpcs with a base of a unique NPC. This caused false positive if other fields, such as an animal's owner, reference a unique NPC.

An example of this in vanilla is Bjorlam, who has exactly one reference but owns his carriage's horse.


Noggog (approved):
Might be good to toss in a unit test flexing the improvement/distinction

kieranknowles1:
I've pushed another commit that adds test cases. This required creating `ContextualRecordTestFixture` as this analyzer relies on mutliple records to function.

Additionally fixes edge case of unique flagged NPCs without the `ActorTypeNPC` keyword (creature races) not being checked.

kieranknowles1:
CI is currently failing on `DetectsExteriorCellWithEncounterZone`, same as the main branch. I've opened #564 to resolve the issue.

PR #567, opened 2026-04-24
Fix missing assets analyzers

kieranknowles1:
Missing assets analysers were always failing due to:
1. They used the implicit (string)File rather than File.DataRelativePath, which would miss the meshes/textures prefix
2. MissingAssetsAnalyzerUtil searched in the executable's working directory rather than data directory

Test cases were passing as they provide an absolute path in AssetLink.GivenPath, which is returned by (string)AssetLink. In real data, GivenPath is relative to Meshes/Textures.

The issue of test case asset links differing from real data is with Mutagen, I am unsure of how to resolve this.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -18,12 +18,12 @@
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="9.0.9" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="9.0.9" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.14.1" />
-    <PackageVersion Include="Mutagen.Bethesda" Version="0.54.0-alpha.22" />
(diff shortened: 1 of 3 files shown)


kieranknowles1:
Tests are failing as the existingModelPath is placed in `C:\\Data\\Meshes\\` while the data directory is set to `C:\GameDirectory\DataDirectory`

Noggog:
PR mentioned is included in Mutagen 0.54.0-alpha.83

kieranknowles1:
Tests will pass once Mutagen is updated

Noggog:
It is, you can probably pull `0.54.0-alpha.83` in and update this PR to point to that version

kieranknowles1:
Another potential issue that’s just came to mind is whether file paths are case sensitive on Mac/linux

Noggog:
Sorry, been AWOL, haha

Hmm, yes they should be, but if the tests line up it shouldn't be a problem?  The tests passing seem like an indication things are lining up

Noggog (approved):

PR #571, opened 2026-04-28
Create analyzer for discussion #570

kieranknowles1:
GetEquipped used with leveled item parameter

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim.Tests/ContextualRecords/ConditionAnalyzerTest.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim.Tests/ContextualRecords/ConditionAnalyzerTest.cs
@@ -1,0 +1,291 @@
+using Mutagen.Bethesda.Analyzers.Skyrim.Record;
+using Mutagen.Bethesda.Analyzers.Testing.Frameworks;
+using Mutagen.Bethesda.Skyrim;
+using Mutagen.Bethesda.Testing.AutoData;
+using Xunit;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.ContextualRecords;
+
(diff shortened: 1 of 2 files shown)


Elscrux:
Where have you encountered issues with this? Do you have any examples?
I'm not sure if this should actually be an analyzer. The conditions were specifically set up to be using IItemOrListGetter as a type because that is what the CK uses. If this is not correct we might want to change the definition in Mutagen instead.

kieranknowles1:
> Where have you encountered issues with this? Do you have any examples? I'm not sure if this should actually be an analyzer. The conditions were specifically set up to be using IItemOrListGetter as a type because that is what the CK uses. If this is not correct we might want to change the definition in Mutagen instead.

IItemOrListGetter is implemented by LeveledItem and the creation kit allows setting one as a parameter, however using a leveled item as a parameter to GetItemCount or GetEquipped will always return 0

The place I initially found this issue was on CYRCloudRulerFF01AlinaDeadScene, in the check for the player having an akaviri katana equipped

Elscrux:
> IItemOrListGetter is implemented by LeveledItem and the creation kit allows setting one as a parameter, however using a leveled item as a parameter to GetItemCount or GetEquipped will always return 0

Yeah exactly, the CK uses a type like `IItemOrListGetter` so it is possible to set it to a leveled list in the CK even though it's wrong.
I think that's a question for @Noggog if the Mutagen api should reflect exactly what the CK allows or if it should be more restrictive and only allow the actual valid options (so no leveled lists). The question then of course becomes how we'd handle if leveled lists were set by the CK. Should there be a Mutagen level thing to handle this or should be simply add an analyzer to tell you that this is actually not a valid configuration for the condition when you set a leveled list as the item.

Noggog:
I think Mutagen should continue to allow for what the CK allows.    I think an analyzer to point out it's sus is a perfect application.  It's allowed, but sus

Noggog (approved):

PR #576, opened 2026-05-02
Add missing temper analysers

kieranknowles1:
Covers discussions #572, #573, #574, #575.

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim/Extensions/ConstructibleExtensions.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/Extensions/ConstructibleExtensions.cs
@@ -2,15 +2,15 @@ using Mutagen.Bethesda.Plugins;
 using Mutagen.Bethesda.Plugins.Cache;
 using Mutagen.Bethesda.Skyrim;
 
-namespace Mutagen.Bethesda.Analyzers.Skyrim.Util;
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Extensions;
 
-public static class TemperRecipeAnalyzerUtil
+public static class ConstructibleExtensions
 {
     public static IEnumerable<IConstructibleObjectGetter> GetTemperRecipes(
+        this IConstructibleGetter item,
         IFormLink<IKeywordGetter> temperKeyword,
         ILinkUsageCache usageCache,
-        ILinkCache linkCache,
-        IConstructibleGetter item)
+        ILinkCache linkCache)
     {
         return usageCache
             .GetUsagesOf<IConstructibleObjectGetter>(item).UsageLinks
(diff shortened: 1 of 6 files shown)


Mutagen.Bethesda.Analyzers.Skyrim/Extensions/ConstructibleExtensions.cs
```diff
@@ -0,0 +1,21 @@
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Plugins.Cache;
+using Mutagen.Bethesda.Skyrim;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Util;
+
+public static class TemperRecipeAnalyzerUtil
+{
+    public static IEnumerable<IConstructibleObjectGetter> GetTemperRecipes(
```
Elscrux, line 9: This function could be changed to be an extension function for `IConstructibleGetter` so you get the option to get the option to resolve the temper recipes for any armor/weapon simply when checking the IntelliSense on the variable. This is definitely also useful outside of this test case.

Extension functions written in the analyzer project should also be migrated to the main Mutagen project over time, then every Mutagen user will get this convenience call.

Elscrux (changes requested):
Nice thought! I have mainly one point that should be changed: the analyzers should filter out enchanted items. Those should never be an option to create yourself pretty much always.

When running it on vanilla that shows that this change removes a lot of the false positives for vanilla.

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/ConstructibleExtensions.cs
kieranknowles1: Changes applied in latest push

kieranknowles1:
> Nice thought! I have mainly one point that should be changed: the analyzers should filter out enchanted items. Those should never be an option to create yourself pretty much always.
> 
> When running it on vanilla that shows that this change removes a lot of the false positives for vanilla.

Not sure I agree on this. Generic enchanted items with templates inherit from their template, which the analyzer already accounts for. Most quest rewards, including those with enchantments, have temper recipes. UESP lists the exceptions to this trend as bugs, see examples below.

https://en.uesp.net/wiki/Skyrim:Drainheart_Sword
https://en.uesp.net/wiki/Skyrim:Shield_of_Solitude

Elscrux:
Ah yes, this is about temper recipes not crafting recipes, that's right

Mutagen.Bethesda.Analyzers.Skyrim/Record/Armor/TemperAnalyzerArmor.cs
```diff
@@ -0,0 +1,63 @@
...
+            572,
+            "No armor temper recipe",
+            Severity.Suggestion)
+        .WithoutFormatting("Armor has no tempering recipe");
+
+    public static readonly TopicDefinition MultipleTemper = MutagenTopicBuilder.FromDiscussion(
+            573,
+            "Multiple armor temper recipies",
+            Severity.Warning)
+        .WithoutFormatting("Armor has multiple tempering recipies");
```
Elscrux: Could you fix all the types with recipies instead of recipes? I think then we're good to go

Elscrux (changes requested):

Elscrux (approved):
Thanks a lot!

PR #579, opened 2026-05-20
 Use usage cache for IsLocationAppliedToInterior and RefTypeSettlementHouseAnalyzer

kieranknowles1:
This is an order of magnitude faster than the previous implementation
that iterated all cells for each location. Replace O(CellsInLoadOrder) iteration with O(1) lookup


Elscrux (approved):
Thanks a lot!

PR #580, opened 2026-05-20
Optimise NoParentLocationAnalyzer and add tests

kieranknowles1:
Tweak order of operations to perform expensive checks last and use usage cache for for worldspace locations.


Elscrux (approved):
Thanks!

PR #581, opened 2026-05-20
Use usage cache for NoMerchantChestLocRefTypeAnalyzerTest

kieranknowles1:
This is an order of magnitude faster than before. Replaces O(FactionsInLoadOrder) iteration with O(1) lookup


Elscrux (approved):
Thanks!

PR #583, opened 2026-05-26
Add analyzer for discussion #582 (Invalid link)

kieranknowles1:
Analyze form links that are not null but cannot be resolved.

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim/Record/LinkAnalyzer.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/Record/LinkAnalyzer.cs
@@ -1,0 +1,61 @@
+using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
+using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Plugins;
+using Mutagen.Bethesda.Plugins.Cache;
+using Mutagen.Bethesda.Skyrim;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Record;
+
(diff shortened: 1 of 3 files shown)


Mutagen.Bethesda.Analyzers.Skyrim/Record/LinkAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public void AnalyzeRecord(ContextualRecordAnalyzerParams<ISkyrimMajorRecordGetter> param)
+    {
+        var record = param.Record;
+
+        foreach (var link in record.EnumerateFormLinks(iterateNestedRecords: false))
+        {
+            // Mutagen doesn't provide enough context to know if a null link is valid
+            if (link.IsNull)
```
Elscrux, line 47: There's the type of nullable form links, so I think we should be able to test that actually.

Elscrux (commented):
This is a great idea to add this kind of analyzer. Thanks!

Mutagen.Bethesda.Analyzers.Skyrim/Record/LinkAnalyzer.cs
kieranknowles1: FormLinkNullable represents how nulls are stored on disk, in many cases it is valid for a regular link to be null such as a linked reference's keyword, a VM adapter's properties, and a cell's lighting template

Elscrux: Yeah, that's true thanks for testing it :)

Elscrux (approved):

PR #594, opened 2026-05-29
Add empty fragment and duplicate fragment analysers

kieranknowles1:
Analysers to detect duplicate fragment function names and empty fragment scripts for dialog responses, packages, perks, quests, and scenes. Covers discussions #584, #585, #586, #587, #588, #589, #590, #591, #592, and #593


Elscrux (approved):
Thanks!

PR #596, opened 2026-05-30
Lod analyzer

Elscrux:
And some extra tweaks

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
+++ b/Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
@@ -10,8 +10,10 @@ using Mutagen.Bethesda.Analyzers.Engines;
 using Mutagen.Bethesda.Analyzers.SDK;
 using Mutagen.Bethesda.Analyzers.SDK.Topics;
 using Mutagen.Bethesda.Environments;
+using Mutagen.Bethesda.Environments.DI;
 using Mutagen.Bethesda.Plugins;
 using Mutagen.Bethesda.Plugins.Order;
+using Mutagen.Bethesda.Plugins.Order.DI;
 using Mutagen.Bethesda.Plugins.Records;
 using Mutagen.Bethesda.Skyrim;
 using Noggog;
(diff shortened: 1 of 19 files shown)


Noggog:
Test failures look like they were introduced in the PR?  Didnt take too close a look at them

Elscrux:
> Test failures look like they were introduced in the PR? Didnt take too close a look at them

Yes, I think it's some issue with my commit to revert handling null plugin.txt's. I did that here https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/pull/596/changes/6f6413808ad58cdf4be4f3aa6944ff58ddf7a921 
The problem was that analyzers wouldn't run in an environment where no plugin listings file is available. I haven't had a chance to look into it more closely yet though :/

Elscrux:
@Noggog I figured out the issue, it's a caused by the DI setup changes I did but actually it's a problem with Mutagen's ability to infer the GameCategory for a more or less empty load order which I fixed here https://github.com/Mutagen-Modding/Mutagen/pull/668.

Could you please get the other PR merged so this one can be updated to pass the tests?

Noggog:
Done

Elscrux:
@Noggog After the Mutagen update, we're passing the tests now 🥳

Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
```diff
@@ -76,34 +78,19 @@ private static IContainer GetContainer(RunAnalyzersCommand command)
...
-                .WithLoadOrder(loadOrder.ToArray());
-
-            if (command.DataFolder is not null)
-            {
-                envBuilder = envBuilder.WithTargetDataFolder(new DirectoryPath(command.DataFolder));
-            }
-
-            gameEnvironment = envBuilder.Build();
-        }
+        builder
```
Noggog, line 81: Seems unrelated to the PR overall?  What was the reasoning for changing things here?  How do the command fields get set in the environment object now?

Mutagen.Bethesda.Analyzers.Cli/Overrides/EmptyPluginListingsPathProvider.cs
```diff
@@ -5,5 +5,5 @@ namespace Mutagen.Bethesda.Analyzers.Cli.Overrides;
 
 internal class NullPluginListingsPathProvider : IPluginListingsPathProvider
 {
-    public FilePath? Get(GameRelease release) => null;
+    public FilePath? Get(GameRelease release) => string.Empty;
```
Noggog, line 8: What needed this change?   Seems like the class is lying now

Mutagen.Bethesda.Analyzers.Cli/RunAnalyzers.cs
Elscrux: Sorry yes I mingled it into this PR, see the reason and specific changes here https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/pull/596/changes/6f6413808ad58cdf4be4f3aa6944ff58ddf7a921
Basically this setup fixes this that the analyzer framework was not able to run in an environment without a Skyrim install which is required for CI applications

Mutagen.Bethesda.Analyzers.Cli/Overrides/EmptyPluginListingsPathProvider.cs
Elscrux: I can rename it to EmptyPluginListingsPathProvider? This was also required in this update that I reverted which broke CI setups https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/commit/6f6413808ad58cdf4be4f3aa6944ff58ddf7a921

Noggog: Hmm, i'd be curious to know why null was a problem but empty string is fine.  Both situations the plugin file doesn't exist

Elscrux: I can't quite remember but it was some Mutagen internals thing. Hard to track down 😅

Noggog: Ah, i think null throws an exception when it goes to read
Empty string does a file.exists check -> no, and then just returns empty plugins
Sort of odd behavior, i might change the logic to throw if it's an empty string too.  That logic is supposed to be for when a legit plugin path simply doesn't exist

I think what we might want for our scenario is to inject an empty listings provider, rather than injecting a path provider

Noggog: I guess it's fine for now. Let's just do that rename, i guess

Elscrux: Yeah, I mean it was that way before also. The change from string.Empty to null only came recently in a refactoring of yours. Idk what the specific purpose but this was the commit https://github.com/Mutagen-Modding/Mutagen.Bethesda.Analyzers/commit/ebe29277939b6105fdcbcafbe935e3f61b2cdbd0

Updated the class name(s) now, another class was still named Null since 2024 or so :D

PR #598, opened 2026-06-04
Create Analyzer for Discussion #597

FaffiSalus:
Created Analyzer to check FormLists for Placed Records

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim.Tests/ContextualRecords/FormLists/FormListWithPlacedRecordAnalyzerTest.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim.Tests/ContextualRecords/FormLists/FormListWithPlacedRecordAnalyzerTest.cs
@@ -1,0 +1,35 @@
+using Mutagen.Bethesda.Analyzers.Skyrim.Record.FormList;
+using Mutagen.Bethesda.Analyzers.Testing.Frameworks;
+using Mutagen.Bethesda.Skyrim;
+using Mutagen.Bethesda.Testing.AutoData;
+using Xunit;
+
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.ContextualRecords.FormLists;
+
+using Fixture = ContextualRecordTestFixture<FormListWithPlacedRecordAnalyzer, FormList, IFormListGetter>;
+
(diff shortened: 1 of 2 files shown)


Mutagen.Bethesda.Analyzers.Skyrim.Tests/ContextualRecords/FormLists/FormListWithPlacedRecordAnalyzerTest.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Mutagen.Bethesda.Analyzers.Skyrim.Tests.ContextualRecords.FormLists;
+
+using Fixture = ContextualRecordTestFixture<FormListWithPlacedRecordAnalyzer, FormList, IFormListGetter>;
+
+public class FormListWithPlacedRecordAnalyzerTest
+{
+    [Theory, MutagenModAutoData]
+    public void DetectFormListWithIPlacedRecord(Fixture fixture)
+    {
+        var cell = fixture.Create<Cell>();
```
Noggog, line 16: I would recommend trying to tweak a few things to clean up this test.

[Theory] + AutoData means you should be able to just request these as parameters on the test directly, rather than calling .Create on the fixture yourself.

If that's not working, it's probably because this test has chosen `MutagenModAutoData`.   This one is meant with the intention of making say, a SkyrimMod, and then requested FormList would be automatically added to it.

What you probably want instead is just the plain old `MutagenAutoData`, which will just make standalone records not added to a mod (like what you're doing manually here).

Noggog (changes requested):

Mutagen.Bethesda.Analyzers.Skyrim.Tests/ContextualRecords/FormLists/FormListWithPlacedRecordAnalyzerTest.cs
FaffiSalus: tbh i kinda just did what i saw in the UniqueNpcPlacementAnalyzerTest 
i'm afraid I don't quite understand what you mean by "you should be able to just request these as parameters", can you point me to an example?

FaffiSalus:
Analyzer actually not needed

PR #603, opened 2026-06-09
Add analyzer for discussion #602

kieranknowles1:
GetRace without vampire race.

Vampires will not be detected for dialogue etc. if there is no check for XRaceVampire

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
@@ -1,5 +1,6 @@
 using Mutagen.Bethesda.Analyzers.SDK.Analyzers;
 using Mutagen.Bethesda.Analyzers.SDK.Topics;
+using Mutagen.Bethesda.Plugins;
 using Mutagen.Bethesda.Skyrim;
 
 namespace Mutagen.Bethesda.Analyzers.Skyrim.Record;
@@ -48,6 +49,11 @@ public class ConditionAnalyzer : IContextualRecordAnalyzer<ISkyrimMajorRecordGet
             Severity.Error)
         .WithFormatting<ILeveledItemGetter>("Condition used with leveled item {0} as parameter");
 
+    public static readonly TopicDefinition<IConditionGetter, IFormLinkGetter<IRaceGetter>> NoVampireRace = MutagenTopicBuilder.DevelopmentTopic(
+            "No vampire condition",
+            Severity.Warning)
+        .WithFormatting<IConditionGetter, IFormLinkGetter<IRaceGetter>>("Condition {0} checks for mortal race {1} but not its vampire equivalent");
+
     public IEnumerable<TopicDefinition> Topics { get; } =
     [
         InvalidConditionReference,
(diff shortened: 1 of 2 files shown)


Elscrux:
@kieranknowles1 Thanks for tackling this!

We'll have to be a bit more fine-grained with the conditions here I fear.

This condition is usually supposed to be set up like this:
GetIsRace NordRace == 1 OR
GetIsRae NordRaceVampire == 1

It's important that both are connected via or, the order doesn't matter.

To properly test this I think it would be good to first create some more utility functions to be able to check this properly.
We should create some function to split a list of conditions into its OR blocks (that is because OR has precedence over AND, and there are no parentheses, otherwise that would be more complex of course) which you can then use to check if the current OR block also contains.

Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
```diff
@@ -57,8 +64,42 @@ public class ConditionAnalyzer : IContextualRecordAnalyzer<ISkyrimMajorRecordGet
         GetCurrentTimeConditionWithAndOnDayBreak,
         GetCrimeGoldRunOnPlayer,
         LeveledItemParameter,
+        NoVampireRace,
     ];
 
+    private static readonly Dictionary<IFormLinkGetter<IRaceGetter>, IFormLinkGetter<IRaceGetter>> VampireRaceLookup = new()
+    {
+        { FormKeys.SkyrimSE.Skyrim.Race.ArgonianRace, FormKeys.SkyrimSE.Skyrim.Race.ArgonianRaceVampire },
+        { FormKeys.SkyrimSE.Skyrim.Race.BretonRace, FormKeys.SkyrimSE.Skyrim.Race.BretonRaceVampire },
```
Noggog: Unrelated trick:

Can put something like
`using SomeNickname = FormKeys.SkyrimSE.Skyrim;`

Then can
`SomeNickname.Race.BretonRaceVampire`

Makes things less verbose in situations like these

Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
```diff
@@ -57,121 +66,186 @@ public class ConditionAnalyzer : IContextualRecordAnalyzer<ISkyrimMajorRecordGet
...
-                    when getItemCount.ItemOrList.Link.TryResolve<ILeveledItemGetter>(param.LinkCache, out var leveledItem):
-                    param.AddTopic(LeveledItemParameter.Format(leveledItem));
-                    break;
-                case IGetEquippedConditionDataGetter getEquipped
-                    when getEquipped.ItemOrList.Link.TryResolve<ILeveledItemGetter>(param.LinkCache, out var leveledItem):
-                    param.AddTopic(LeveledItemParameter.Format(leveledItem));
-                    break;
+            foreach (var orBlock in conditions.SplitOrBlocks())
+            {
+                foreach (var condition in block)
```
Elscrux: Should this block that is enumerated here not be orBlock? It seems like it should use that, otherwise we'll try to detect for every GetIsRace condition if all or blocks have the vampire variants, and not just the current one?

Mutagen.Bethesda.Analyzers.Skyrim/Extensions/RecordExtension.cs
```diff
@@ -4,36 +4,41 @@ namespace Mutagen.Bethesda.Analyzers.Skyrim.Extensions;
 
 public static class RecordExtension
 {
-    public static IEnumerable<IConditionGetter>? GetConditions(this ISkyrimMajorRecordGetter record)
+    /// <summary>
+    /// Get conditions attached to the record, split by field
+    /// </summary>
+    /// <param name="record"></param>
+    /// <returns></returns>
+    public static IEnumerable<IEnumerable<IConditionGetter>> GetConditions(this ISkyrimMajorRecordGetter record)
```
Noggog: Maybe add `ByField` to the name

Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
```diff
@@ -57,121 +66,206 @@ public class ConditionAnalyzer : IContextualRecordAnalyzer<ISkyrimMajorRecordGet
         GetCurrentTimeConditionWithAndOnDayBreak,
         GetCrimeGoldRunOnPlayer,
         LeveledItemParameter,
+        NoVampireRace,
     ];
 
-    public void AnalyzeRecord(ContextualRecordAnalyzerParams<ISkyrimMajorRecordGetter> param)
+    private static readonly Dictionary<IFormLinkGetter<IRaceGetter>, IFormLinkGetter<IRaceGetter>> VampireRaceLookup = new()
```
Noggog, line 72: I think all this new logic is piling into ConditionAnalyzer too much.  Already was a past sin of the father, but this is just perhaps the last straw for the camel.

If we added all logic we ever wanted to do relating to conditions into this file, it would become unmanagable.

we should split the vampire checks off into a new analyzer.  And I'd argue all other topics should be split up into their own similar files so that the umbrella "ConditionAnalyzer" no longer exists at all

Elscrux: Yeah this has to be split at some point, probably good to do it now for sure. I didn't do that from the start because I feared that calling GetConditions on all major records not just once but as many times as there are condition analyzers will create a bottleneck but it's probably quite fine.

What would be huge for this would be to have some kind of aspect interface in Mutagen which we can add to pre-filter records with conditions. Those conditions are not in a shared link of course (see the conditions retrieval function), but it would still be great to filter this more.
Also those helper function to get all the conditions from some record (as many others that we added in analyzers) should be moved to Mutagen so everybody can enjoy them.

Noggog: I think for a lot of them we can refactor to an IndividualConditionAnalyzer

Make the ConditionAnalyzer inject all the IndividualConditionAnalyzers and loop them.  This allows the logic for handling each live elsewhere, while still looping them once.  

But yeah, probably for another PR, hah

PR #607, opened 2026-06-11
Create analyzer for discussion #606 landscape heightmap seam

kieranknowles1:
Discussions #606 landscape heightmap seam and #609 vertex colour seam


kieranknowles1:
Keeping this as a draft until test cases are added

Mutagen.Bethesda.Analyzers.Skyrim/Record/Landscape/LandscapeSeamAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+        void CheckSeams<T>(TopicDefinition<Direction> topic, Func<ILandscapeGetter, IReadOnlyArray2d<T>?> getData)
+            where T : IEquatable<T>
+        {
+            var data = getData(landscape);
+            if (data == null)
+                return;
+
+            void CheckNeigbour(Direction dir)
+            {
+                // TODO: Worldspace.GetCell is fairly slow.
```
Elscrux: If you have ideas how to speed up the call that'd be great! We could also check if @Noggog has any ideas.

One thing that would help for this analyzer at least is to only get each neighboring cell once. Currently we get it twice, once for the heightmp and once for the vertex colors. Though if Mutagen is the performance bottleneck then this won't help much I assume because it caches its results anyway and then it's just on getting the data for the first time really.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Landscape/LandscapeSeamAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+                    param.AddTopic(topic.Format(dir));
+            }
+            CheckNeigbour(Direction.North);
+            CheckNeigbour(Direction.East);
+            CheckNeigbour(Direction.South);
+            CheckNeigbour(Direction.West);
+        }
+
+        CheckSeams(HeightMapSeam, l => l.VertexHeightMap?.Decode());
+        CheckSeams(VertexColorSeam, l => l.VertexColors ?? DefaultVertexColors);
```
Elscrux, line 148: It would also be GREAT to get a check for landscape textures here. It's a bit more involved because the data is stored in a slightly different way but still. This is not just between neighboring cells btw, but also it's a thing within cells because landscape textures are applied in quads within cells. Could still fit in this analyzer.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Landscape/LandscapeSeamAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    static bool HasSeam<T>(IEnumerable<T> a, IEnumerable<T> b)
+        where T : IEquatable<T>
+    {
+        foreach (var (a1, b1) in a.Zip(b))
+        {
+            // Don't need a large epsilon here. While the heightmap is a float, all vanilla landscape uses integer offsets
+            if (!a1.Equals(b1))
+                return true;
```
Elscrux: I'd be great to count how many deviations there are and report that. That would make it easier to know how much needs to be fixed.
Each individual deviation and the difference could even be passed into the meta data of the result.

Mutagen.Bethesda.Analyzers.Skyrim/Record/Landscape/LandscapeSeamAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    public void AnalyzeRecord(ContextualRecordAnalyzerParams<ILandscapeGetter> param)
+    {
+        var landscape = param.Record;
+
+
+        var cell = landscape.GetCell(param.LinkCache);
+        var worldspace = cell?.GetWorldspace(param.LinkCache);
+        if (cell?.Grid == null || worldspace == null) return;
```
Elscrux, line 106: I'm currently getting various reports in vanilla outside the playable area which I think we should filter out. We could use the border region for this distinction if a cell is relevant or not. Or the border region + 2 cells outside of it in case you can still see outside of it.

This whole thing should probably be in an extension function.

Elscrux (changes requested):
Very cool to see this. I have some feedback to improve this.

Mutagen.Bethesda.Analyzers.Testing/Frameworks/ContextualRecordTestFixture.cs
```diff
@@ -52,7 +52,7 @@ ContextualRecordAnalyzerParams<TMajorGetter> CreateAnalyserParams(TestParameters
 
     TestParameters Setup()
     {
-        var mod = new SkyrimMod(ModKey.Null, SkyrimRelease.SkyrimSE);
+        var mod = new SkyrimMod("mod1.esp", SkyrimRelease.SkyrimSE);
```
Noggog: In situations like this where the value doesnt matter (could be "Derp.esp" and work just the same), I prefer leaning on the fixture to provide things.   Can just request a ModKey to give it here.   Helps clarify what's important to a test vs just filler data

Mutagen.Bethesda.Analyzers.Skyrim.Tests/ContextualRecords/Landscapes/LandscapeSeamAnalyzerTest.cs
```diff
@@ -0,0 +1,107 @@
...
+    {
+        using var dataMod = GetTestingPlugin();
+        using var linkCache = dataMod.ToImmutableLinkCache();
+
+        fixture.Run(
+            prepForError: (rec, mod) =>
+            {
+                var southLand = Setup(rec, mod);
+
+                rec.VertexHeightMap = linkCache.Resolve<ILandscapeGetter>(TestLandBad).VertexHeightMap!.DeepCopy();
```
Noggog: Feels like we just need to add to the testing suite to help facilitate tests against files a bit more.

This test fixture.Run is meant for records were making in memory straight and then testing.

You've sidestepped it to work with the file sourcing here, but we could probably make a similar setup that was more of a first class citizen in the testing framework

I'll think on it a bit

Mutagen.Bethesda.Analyzers.Skyrim/Record/Landscape/LandscapeSeamAnalyzer.cs
Noggog: I think the answer is introducing some sort of IWorldspaceCellLocator interface.  It would be a wrapper around ILinkInterface that does some additional caching for the purpose of looking up cells like this.   Once designed, I think we had some plans/mechanisms for how custom objects might be injected/reused in analyzers?   I'd have to brush up

EDIT:
Ah yeah, we have the IProvideCaches interface on the params already
```cs
public void AnalyzeRecord(ContextualRecordAnalyzerParams<ILandscapeGetter> param)
{
    var worldspaceCellsCache = param.ResolveCache<IWorldspaceCellLocator>();
    // Use it
}
```


Just need to design the implementation of WorlspaceCellLocator to cache and reuse results as best it can

Mutagen.Bethesda.Analyzers.Testing/Frameworks/ContextualRecordTestFixture.cs
kieranknowles1: Changing the modkey to something other than null is necessary here as the link cache doesn’t resolve records with ModKey.Null

This could be changed to the mod being provided by the fixture which would allow using Mod.AddNew rather than Fixture.Create and automatically insert fixture generated records into the mod

Mutagen.Bethesda.Analyzers.Skyrim.Tests/ContextualRecords/Landscapes/LandscapeSeamAnalyzerTest.cs
kieranknowles1: File sourcing is used here primarily since I found it easier than generating a height map in the test case

Elscrux:
Is this cell cache something that would be relevant for Mutagen too? Seeing as cell lookup performance issues aren't just a problem for the analyzers but all Mutagen programs.

kieranknowles1:
Added support for landscape texture seams. One consideration I've had here is the potential of grouping textures by texture set, disregarding grass. There would still be a seam between grass types, but no seam between textures

Elscrux:
> Added support for landscape texture seams. One consideration I've had here is the potential of grouping textures by texture set, disregarding grass. There would still be a seam between grass types, but no seam between textures

Great job! I'll check it out in the coming days. The grouping idea is good, maybe we want to have two topics, one for not matching textures in general, and one for grass vs no grass (or different grass)?

PR #608, opened 2026-06-12
ConsoleReportHandler reports EditorID of sourceMod instead of Winning EditorID of LinkCache

FaffiSalus:
ConsoleReportHandler reports EditorID of sourceMod instead of Winning EditorID of LinkCache

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Engine/Reporting/Handlers/ConsoleReportHandler.cs
+++ b/Mutagen.Bethesda.Analyzers.Engine/Reporting/Handlers/ConsoleReportHandler.cs
@@ -3,6 +3,7 @@ using Mutagen.Bethesda.Analyzers.SDK.Drops;
 using Mutagen.Bethesda.Analyzers.SDK.Topics;
 using Mutagen.Bethesda.Plugins;
 using Mutagen.Bethesda.Plugins.Cache;
+using Mutagen.Bethesda.Plugins.Cache.Internals.Implementations;
 using Mutagen.Bethesda.Plugins.Records;
 using Noggog.WorkEngine;
 
@@ -33,7 +34,24 @@ public class ConsoleReportHandler : IReportHandler
         _workDropoff.Enqueue(() =>
         {
             IMajorRecordGetter? parent = null;
-            _linkCache.TryResolveIdentifier(majorRecord, out var editorId);
+            ImmutableModLinkCache? sourceModLinkCache = null;
+            string? editorId = "";
+            foreach (var modGetter in _linkCache.ListedOrder)
+            {
+                if (modGetter.ModKey == sourceMod)
+                {
+                    sourceModLinkCache = new ImmutableModLinkCache(modGetter);
+                }
+            }
+            if (sourceModLinkCache is null)
+            {
+                _linkCache.TryResolveIdentifier(majorRecord, out editorId);
+            }
+            else
+            {
+                sourceModLinkCache.TryResolveIdentifier(majorRecord, out editorId);
+            }
+
             if (_linkCache.TryResolveSimpleContext(majorRecord, out var parentContext) && parentContext?.Parent?.Record is IMajorRecordGetter parentRecord)
             {
                 parent = parentRecord;
--- a/Mutagen.Bethesda.Analyzers.Engine/Reporting/Handlers/ConsoleReportHandler.cs
+++ b/Mutagen.Bethesda.Analyzers.Engine/Reporting/Handlers/ConsoleReportHandler.cs
@@ -34,22 +34,16 @@ public class ConsoleReportHandler : IReportHandler
         _workDropoff.Enqueue(() =>
         {
             IMajorRecordGetter? parent = null;
-            ImmutableModLinkCache? sourceModLinkCache = null;
-            string? editorId = "";
-            foreach (var modGetter in _linkCache.ListedOrder)
+            var editorId = "";
+            var modGetter = _linkCache.ListedOrder.FirstOrDefault(l => l.ModKey == sourceMod);
+            if (modGetter is not null)
             {
-                if (modGetter.ModKey == sourceMod)
-                {
-                    sourceModLinkCache = new ImmutableModLinkCache(modGetter);
-                }
-            }
-            if (sourceModLinkCache is null)
-            {
-                _linkCache.TryResolveIdentifier(majorRecord, out editorId);
+                var sourceModLinkCache = new ImmutableModLinkCache(modGetter);
+                sourceModLinkCache.TryResolveIdentifier(majorRecord, out editorId);
             }
             else
             {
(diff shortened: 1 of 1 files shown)


Mutagen.Bethesda.Analyzers.Engine/Reporting/Handlers/ConsoleReportHandler.cs
```diff
@@ -33,7 +34,24 @@ public void Dropoff(
         _workDropoff.Enqueue(() =>
         {
             IMajorRecordGetter? parent = null;
-            _linkCache.TryResolveIdentifier(majorRecord, out var editorId);
+            ImmutableModLinkCache? sourceModLinkCache = null;
+            string? editorId = "";
+            foreach (var modGetter in _linkCache.ListedOrder)
```
Elscrux: In C# we can avoid a lot of loops like these with LINQ, and you would use something like FirstOrDefault() here to get the mod that you are searching for and based on if it returns null or not you would then create the link cache for the mod.

Mutagen.Bethesda.Analyzers.Engine/Reporting/Handlers/ConsoleReportHandler.cs
```diff
@@ -33,7 +34,24 @@ public void Dropoff(
...
+            {
+                if (modGetter.ModKey == sourceMod)
+                {
+                    sourceModLinkCache = new ImmutableModLinkCache(modGetter);
+                }
+            }
+            if (sourceModLinkCache is null)
```
Elscrux: Though I think creating a new link cache each time you drop off any report that's not very efficient because you basically have to construct and fill the cache all the time and throw it away, only using it once. Instead what we would like to use is a call like this ResolveAllSimpleContexts which gives you a list of all the record overrides and the matching mod that they're part of then you can see if there's an override from your source mod and then use that EditorID.

Elscrux: @Noggog Is this (getting a specific override from a certain mod) something we should add to Mutagen as a default call?

Elscrux (changes requested):
Thanks, that's a good thing to we need to be sure to fix, as a lot of the data we use currently is the info from the winning override. Maybe there are also other points where this would be necessary.

Mutagen.Bethesda.Analyzers.Engine/Reporting/Handlers/ConsoleReportHandler.cs
FaffiSalus: I've tried using _linkCache.ResolveAllSimpleContexts<MajorRecord>(majorRecord.FormKey) but i get no results from it, what am I doing wrong here?

Elscrux: Strange, that should definitely work, though you should best use a version of the function where you can pass a type (your IDE should tell you that the ResolveAllSimpleContexts version not with any type specified via additional argument or generics is very slow). I'd assume that also the slow version would give you the results though.

FaffiSalus: I used this _linkCache.ResolveAllSimpleContexts<MajorRecord>(majorRecord.FormKey).Count() and printed out the result to test it and I keep coming up with 0, do you want to test it in your environment? The version with a fresh ImmutableModLinkCache works but as you said it's probably not the best for performance

Noggog:
Hmm, so the goal is just to prefer the EditorID as the record originally defined it, rather than the winning override which can sometimes change the EditorID?   For one, I'd like to hear why that's more desired.  Just preference?  Or some other reasons?

I think the current code can probably be swapped to just make use of the resolve targets, [described here](https://mutagen-modding.github.io/Mutagen/linkcache/#resolve-target).   Should just be an extra parameter on the original TryResolveIdentifier call

Elscrux:
> Hmm, so the goal is just to prefer the EditorID as the record originally defined it, rather than the winning override which can sometimes change the EditorID? For one, I'd like to hear why that's more desired. Just preference? Or some other reasons?

It is needed because it makes it super confusing when you report that a record with EditorID "X" is bugged, but then you actually meant the override of the record that is called "A". You would assume that the override "X" is bugged even though the report was for "A". I don't think there is a reason that we shouldn't report the EditorID of the actual override that we report something about.

> I think the current code can probably be swapped to just make use of the resolve targets, [described here](https://mutagen-modding.github.io/Mutagen/linkcache/#resolve-target). Should just be an extra parameter on the original TryResolveIdentifier call

I don't think this works. The resolve targets are just for checking if you want to get the first override of a record or the last (winning) override. But what we want is the actual override from a certain mod (and that was the question for you if we can get a call like that in Mutagen). Because the order of overrides might be A, B, and C. And we want the override B, not one at the start or end.

FaffiSalus:
So the way it currently works is the Analyzer analyzes Mod A and reports "there is an issue in Mod A" but it reports this issue with the EditorID from Mod B. 
While you still can find the correct record with this Information it's not quite correct and is confusing for users.
I think Analyzers should only give out the information that is relevant to the mod that is currently being analyzed.

Noggog:
Ah gotcha.  So if there's 15 overrides of record A, and the 9th has an error, we want to display the editorID as it is from that 9th version

Noggog:
I think the fix is to expose a component like ILinkCachePerModProvider.   It would be an interface defined by the analyzer engine, so it could fulfill it once as needed and cache as long as applicable.   The interface would let you give it a mod key and get a link cache back that was made just against that one mod to query against.

I'll have to look at if theres ways to optimize its fulfillment to reuse caches of the main link cache or not.  Not a big deal either way

Elscrux:
> I think the fix is to expose a component like ILinkCachePerModProvider. It would be an interface defined by the analyzer engine, so it could fulfill it once as needed and cache as long as applicable. The interface would let you give it a mod key and get a link cache back that was made just against that one mod to query against.
> 
> I'll have to look at if theres ways to optimize its fulfillment to reuse caches of the main link cache or not. Not a big deal either way

That would also fix it but wouldn't it also be good to have a way in Mutagen to get an override from a specific Mod/ModKey directly?

Noggog:
How about this PR?
https://github.com/Mutagen-Modding/Mutagen/pull/670

FaffiSalus:
That looks like a good solution to me, I'll adjust this code once the new functionality is available, thank you

FaffiSalus:
@Noggog @Elscrux 
I adjusted the Code to use the new function and also updated the Nuget packages, please check if I did the Nuget part correctly and whether it needs to go in the PR

Noggog:
I bumped them on `dev`.  Can you reapply/clean this PR after pulling it in?

PR #612, opened 2026-06-16
Add analysers for discussions #610 and #611

kieranknowles1:
Invalid alias index and alias function without owning quest

### DIFF
--- a/Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
+++ b/Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
@@ -48,6 +48,18 @@ public class ConditionAnalyzer : IContextualRecordAnalyzer<ISkyrimMajorRecordGet
             Severity.Error)
         .WithFormatting<ILeveledItemGetter>("Condition used with leveled item {0} as parameter");
 
+    public static readonly TopicDefinition<IConditionGetter> AliasWithoutQuest = MutagenTopicBuilder.FromDiscussion(
+            611,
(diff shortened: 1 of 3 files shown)


Mutagen.Bethesda.Analyzers.Skyrim/Record/ConditionAnalyzer.cs
```diff
@@ -172,6 +200,12 @@ when getItemCount.ItemOrList.Link.TryResolve<ILeveledItemGetter>(param.LinkCache
                     when getEquipped.ItemOrList.Link.TryResolve<ILeveledItemGetter>(param.LinkCache, out var leveledItem):
                     param.AddTopic(LeveledItemParameter.Format(leveledItem));
                     break;
+                case IGetIsAliasRefConditionDataGetter getIsAliasRef:
+                    CheckAlias(getIsAliasRef.ReferenceAliasIndex);
+                    break;
+                case { RunOnType: Condition.RunOnType.QuestAlias }:
```
Elscrux: This should be in a separate if branch, other switch cases might take precedence here and this case might not be running for some scenarios where we'd want it too.

Elscrux (changes requested):
Good stuff! Love the GetOwningQuest() extension method btw :D

We'll just need one small change

Elscrux:
Oh sorry I actually meant that just this case
> case { RunOnType: Condition.RunOnType.QuestAlias }:

needs to be separated, the other one is switching on the data which should be fine. But it works either way

Noggog:
Needs some merge love with the other vampire condition PR merges from earlier

